PR #3, opened 2017-11-15
Task/arxivng 138

eawoods:
Added in the user menu and search bar - these are not yet connected to their respective services, to make it more obvious that this requires some code to interface with the classic system. 

CSS/sass structure has also been refactored to make better use of the Bulma framework.


eawoods:
Ah nuts, that was supposed to be a request to merge into develop.

PR #4, opened 2017-11-15
Task/arxivng 138

eawoods:
Added user menu and search bar to base UI headers. Note that there is code missing from both to activate the form and related links - to make it more obvious that interface is needed from other parts of the system.

CSS/sass refactoring has also been done to better take advantage of the Bulma framework.


baseui/templates/baseui/base.html
```diff
@@ -36,6 +36,16 @@
     </div>
     <!-- contains arXiv identity and search bar -->
     <div class="identity"><h1><a href="">arXiv.org</a></h1></div>
+    <div class="user-tools">Login | My Account | Logout</div>
+    <div class="search-block">
+      <form>
```
erickpeirson: I hate to advocate hard-coding anything, but the only thing missing here is ``method="POST" action="https://arxiv.org/search"`` ; @mhl10 what do you think?

PR #6, opened 2017-12-01
Task/arxivng 159

eawoods:
- Broke out base template into head, header, and footer segments
- Added macros for pagination (incomplete) and inline search bar
- Added support for MathJax via CDN with inline custom config


baseui/templates/baseui/macros.html
```diff
@@ -0,0 +1,50 @@
+{# macros to be available to all templates across arxiv #}
```
erickpeirson: I like this! As they grow more numerous, would you continue to keep them in one file? Or split them out into separate files?

baseui/templates/baseui/head.html
```diff
@@ -0,0 +1,32 @@
+<meta charset="utf-8">
+<meta name="viewport" content="width=device-width, initial-scale=1">
+<link rel="icon" type="image/png" href="{{ url_for('baseui.static', filename='images/icons/favicon-32x32.png')}}" sizes="32x32" />
+<link rel="icon" type="image/png" href="{{ url_for('baseui.static', filename='images/icons/favicon-16x16.png')}}" sizes="16x16" />
+<title>{{ pagetitle }}</title>
+<link rel="stylesheet" href="https://maxcdn.bootstrapcdn.com/font-awesome/4.7.0/css/font-awesome.min.css" />
+<link rel="stylesheet" href="{{ url_for('baseui.static', filename='css/arxivstyle.css')}}" />
+<script type="text/x-mathjax-config">
```
erickpeirson: I could see breaking out the mathjax tags into a separate template fragment. But not essential.

baseui/templates/baseui/footer.html
```diff
@@ -0,0 +1,20 @@
...
+      <li><a href="https://confluence.cornell.edu/display/arxivpub/arXiv+Public+Wiki">Wiki - project documentation</a></li>
+    </ul>
+  </div>
+  <div class="column">
+    <ul class="nav-spaced">
+      <li><a href="mailto:web-accessibility@cornell.edu">Web Accessibility Help</a></li>
+      <li><a href="/help">Help with using arXiv</a></li>
+    </ul>
+  </div>
+</div>
```
erickpeirson: Seeing quite a few things here that might belong in a configuration file (email addresses, blog URL, etc). Something to think about for the future.

baseui/templates/baseui/macros.html
eawoods: Single vs multiple files: I suppose that depends on how many macros we end up needing. Ideally, we'd have some kind of list of components at this level in advance of the template creation...but we gotta work with what we've got.
I was imagining that we'd have separate macro files per section if the functions are section-specific, but I couldn't think of any and decided that doing it as one file for now was pragmatic.

If the macros start to emerge into groupings, separate files might be easier to manage.

baseui/templates/baseui/footer.html
eawoods: Yes, I agree on the config file - also think that can happen in the future when we know more about what we need to have here (this footer is pretty much untested and isn't quite what we have now).

baseui/templates/baseui/head.html
eawoods: Good point. Kind of hard to draw the line, at this moment, on which pieces are big enough for their own fragment. Perhaps when we're closer to done we can review for sensible pieces?

erickpeirson: Makes sense to me!

PR #7, opened 2017-12-01
ARXIVNG-163 added minimal flask app for verification/testing

erickpeirson:
@eawoods What do you think of this?

### DIFF
--- a/README.md
+++ b/README.md
@@ -13,6 +13,19 @@ arxiv-NG, and can be overridden as needed by submodules.
 testpage.html is included for rapid evaluation of display layer CSS without
 invoking a local environment to render templates.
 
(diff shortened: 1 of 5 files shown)


eawoods:
@erickpeirson looking good! Instead of rendering testpage.html, could you set it to render base.html? Testpage doesn't have any template language in it - once your app is running testpage can be deleted (I have to carry over changes from testpage to base by hand, presently, and only created testpage.html as a way to see what I was doing because I didn't have a means of rendering Jinja).

erickpeirson:
@eawoods How now?

erickpeirson:
Awesome, and no worries!

PR #8, opened 2018-01-19
Task/arxivng 200

eawoods:
Agonized over search button color and may still need to adjust it later. #todo Would like to make a fancy feature to make any color x% darker and retain hover/focus color functions; I'll have to level up my Sass skills for that. Sure that it's a fine point rather than a need-to-have.


erickpeirson:
@eawoods Mind changing base to ``develop``?

eawoods:
*headdesk* done

PR #10, opened 2018-01-25
ARXIVNG-227 added schema validation helpers to arxiv.util

erickpeirson:


erickpeirson:
@JaimieMurdock Docs: agreed. Depend: yes, good idea.

PR #11, opened 2018-01-25
Minor documentation updates due to migration from arxiv-base-ui to ar…

DavidLFielding:
…xiv-base.


erickpeirson:
@DavidLFielding I'll leave this for you to merge when ready; some of us got in the habit of doing this just in case there were minor last-minute changes that should be included.

PR #12, opened 2018-01-29
renamed Dockerfile; support pipenv

bbarker:


bbarker:
Apparently I need to spellcheck my git commit messages

bbarker:
Looks like something is currently going wrong with the ubuntu environment specified for travis - probably a temporary package resolution issue.

PR #16, opened 2018-02-07
ARXIVNG-204 implemented config-driven URL builder for base templates

erickpeirson:
@eawoods This is to address building URLs for things in the base templates. It seems that the biggest constraint on this functionality is that it needs to be possible to make changes to these URLs without touching code (since base will be almost everywhere). The most straightforward way to do this is via configuration. So in this implementation, a function called ``config_url(target: str) -> str`` gets injected into the template context; it looks for and returns a configuration variable called ``ARXIV_[TARGET]_URL``.

### DIFF
--- a/app.py
+++ b/app.py
@@ -1,5 +1,5 @@
 """Provides application for development purposes."""
 
-from base.factory import create_web_app
+from arxiv.base.factory import create_web_app
 
 app = create_web_app()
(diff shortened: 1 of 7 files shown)


erickpeirson:
Yes. To add a new URL, you would add to configy.py:

```python
ARXIV_COOL_URL = os.environ.get('ARXIV_COOL_URL', '/cool/default/url')
```

and then you should be able to use it in the template with:

```html
<a href="{{ config_url('cool') }}">Cool</a>
```

PR #21, opened 2018-02-15
Simple dict-based representation of the arXiv category taxonomy.

mhl10:
This is essentially a conversion of the original Perl code but consolidated in a single file. I'm not sure we need more complexity than this at the moment but welcome any suggestions that could help make our lives easier later.

### DIFF
--- a/arxiv/taxonomy.py
+++ b/arxiv/taxonomy.py
@@ -1,4 +1,5 @@
 """arXiv group, archive and category definitions."""
+from datetime import date
 
 GROUPS = {
     'grp_physics': {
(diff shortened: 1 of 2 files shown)


arxiv/taxonomy.py
```diff
@@ -0,0 +1,1479 @@
...
+    'acc-phys': {
+        'name': 'Accelerator Physics',
+        'in_group': 'grp_physics',
+        'start_date': '1994-11',
+        'end_date': '1996-09',
+    },
+    'adap-org': {
+        'name': 'Adaptation, Noise, and Self-Organizing Systems',
+        'in_group': 'grp_physics',
+        'start_date': '1993-03',
```
erickpeirson: Nit-picky, but any reason not to use ``datetime.date``s for these?

mhl10: Sounds good to me, will update

mhl10:
Now updated with date objects

PR #23, opened 2018-02-28
ARXIVNG-313 added exception handling and templates

erickpeirson:
This still needs work, but let's flag this for discussion.


erickpeirson:
This uses Flask's exception handling pattern to render a standard set of error templates. A service that uses Base will inherit these handlers. Controllers may simply raise  Werkzeug HTTPExceptions, and the rest is taken care of.

One question we should think about is what information will we want to be able to control on the error page when raising an exception. Right now, the error description (the message with which the exception is raised) is rendered in a paragraph element, and can include markup. One thing that come to mind: is a request ID of some kind (included in application logs). What else?

Templates: what else should be included on the base templates? Contact info to report the error? Links to help pages? Skull and crossbones?

eawoods:
Items to include in templates:
- helpful title (maybe along the pattern of "### - <Error Reason>"
- If we want, can use the cute graphic-smiley to illustrate the number (all but 413 have at least one zero and can sub an SVG of that image for zeroes)
- Search box for 404, maybe a few others that make sense
- Link to front page
- Link to contact for error reporting
- Helpful information about how to not get the same error

Seems like a lot, but can be compact. Feel free to assign generating the template content to me, or I can just take it on. ;-)

erickpeirson:
@eawoods Added ARXIVNG-313 ;-)

arxiv/base/exceptions.py
```diff
@@ -1,5 +1,117 @@
...
+from arxiv import status
+
+_handlers = []
+
+
+class ConfigurationError(InternalServerError):
     """Raised when a configuration parameter is missing or invalid."""
+
+
+def handler(exception: type) -> Callable:
```
JaimieMurdock: I like the generic handler decorator - clever architecture.

arxiv/base/routes.py
```diff
@@ -1,14 +1,61 @@
 """Provides routes for verifying base templates."""
```
JaimieMurdock: Would add to docstring note about verifying exceptions too.

arxiv/base/templates/base/401.html
```diff
@@ -0,0 +1,20 @@
+{% extends "base/base.html" %}
+
+{% block content %}
+<section class="hero">
+  <div class="hero-body">
+    <div class="container">
+      <h1 class="title">
+        Four-Oh-One
```
JaimieMurdock: Any reason to spell these out?

mypy.ini
```diff
@@ -4,7 +4,7 @@ mypy_path = $MYPYPATH:./sqlalchemy-stubs
 #
 # Covered by --strict, with some turned off:
 #
-disallow_untyped_calls=True
+#disallow_untyped_calls=True
```
JaimieMurdock: Will also be added by `task/ARXIVNG-293`, so no conflicts there.

arxiv/base/templates/base/401.html
erickpeirson: No reason :-)

arxiv/base/routes.py
erickpeirson: Will do

erickpeirson:
Merging to clear the way for ARXIVNG-314

PR #25, opened 2018-03-06
ARXIVNG-315 accessibility tidy up page title block and compile CSS

eawoods:
corrects accessibility errors for color contrast in footer, search labels in header, page title


arxiv/base/templates/base/head.html
```diff
@@ -2,7 +2,7 @@
 <meta name="viewport" content="width=device-width, initial-scale=1">
 <link rel="icon" type="image/png" href="{{ url_for('base.static', filename='images/icons/favicon-32x32.png')}}" sizes="32x32" />
 <link rel="icon" type="image/png" href="{{ url_for('base.static', filename='images/icons/favicon-16x16.png')}}" sizes="16x16" />
-<title>{{ pagetitle }}</title>
+<title>{% block pagetitle %}arXiv e-print repository{% endblock %}</title>
```
erickpeirson: One thought: what about...

```<title>{{ pagetitle }} | arXiv e-print repository</title>```

So that we can set ``pagetitle="foo title"`` in the template rendering context, resulting in:

``<title>foo title | arXiv e-print repository</title>``

Comment (eawoods) [reply arxiv/base/templates/base/head.html]: Excellent. I was hoping for such a suggestion!  For some reason -- {{ pagetitle }} wasn't rendering at all (perhaps because it's not set anywhere yet?) but I would love to use it.

Comment (erickpeirson) [reply arxiv/base/templates/base/head.html]: https://github.com/cul-it/arxiv-base/blob/6c58e5e669d7e4536e7749eb91396e79ddfaa027/arxiv/base/routes.py#L14

Could do: 

```python
    return render_template("base/base.html", pagetitle="Foo title")
```

eawoods: How does this magic work when we have other pages using the base template? Can the page title then be overridden by child templates in the same fashion?

erickpeirson: Doesn't even require intervention by the child template -- it just needs to be set in the template rendering context. So a route function could set it, or a context processor middleware could set it.

PR #39, opened 2018-03-13
ARXIVNG-339 increase contrast of lightest grey outlines

eawoods:
Changed lightest grey after viewing on high brightness/contrast monitor.
CSS variables FTW.


arxiv/base/static/css/arxivstyle.css
```diff
@@ -555,10 +555,10 @@ a.has-text-danger:hover, a.has-text-danger:focus {
...
 .has-text-grey-light {
-  color: #999 !important; }
+  color: #999999 !important; }
 
 .has-text-grey-lighter {
-  color: #eee !important; }
+  color: #cccccc !important; }
 
 .has-text-white-ter {
   color: whitesmoke !important; }
```
erickpeirson: General question: in the dark recesses of my memory I remember reading an article that claimed that use of !important was smelly (as in code smell). What's your take on that?

eawoods: Yeah, I don't really like it all that much either - but the dark recesses of my own memory says that might have been from back-in-the-day? Bulma uses !important as part of its override-layering mechanism, so we may not have too much control over such things.

erickpeirson: Gotcha. Thanks!

eawoods: Setting a bookmark for myself for some quick research on that, after beta.

PR #41, opened 2018-03-22
Update MathJax config to only target elements with "mathjax" class

mhl10:
All other elements are skipped. Since we haven't introduced abstracts yet, the "mathjax" class has only been applied to the title in the search results.

This PR also introduces two other minor changes:

1. disable multi-line, black-bordered box around TeX could not be processed by MathJax; it now looks like the rest of the paragraph
2. hide grey rendering status box that appears at the bottom of the page on page load; it was distracting to some users and doesn't provide much useful feedback anyway. Note that it's also possible to disable the greying of the TeX elements that MathJax has queued to process, but I haven't made that change here.

### DIFF
--- a/arxiv/base/templates/base/head.html
+++ b/arxiv/base/templates/base/head.html
@@ -23,7 +23,7 @@
           inlineDelimiters: ["$","$"],
           multiLine: false,
           style: {
-            "font-size": "normal"',
+            "font-size": "normal",
             "border": ""
           }
         }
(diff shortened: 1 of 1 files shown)


mhl10:
Hi @eawoods, for testing I've modified my local copy of requirements/dev.txt to point to the branch in arxiv-base I'm trying to test, e.g.:
`` 
#arxiv-base==0.4.1
-e git://github.com/cul-it/arxiv-base.git@task/ARXIVNG-381#egg=arxiv-base
``

eawoods:
Thank you @mhl10!

mhl10:
Thanks @eawoods, will merge!

PR #45, opened 2018-03-30
ARXIVNG-392 ARXIVNG-391 added logging module with reasonable defaults

erickpeirson:
This adds the module ``arxiv.base.logging``, which provides a function ``getLogger`` -- it works just like the builtin ``logging.getLogger``, but with a simpler interface ( ``getLogger(name: str, stream: TextIO)`` ). It sets some defaults that should be applied consistently across all arXiv services, so that we can parse application log messages in a consistent way.

### DIFF
--- a/arxiv/base/logging.py
+++ b/arxiv/base/logging.py
@@ -57,7 +57,7 @@ def getLogger(name: str, stream: IO = sys.stderr) -> logging.Logger:
 
     # Log messages should be in Eastern local time, for consistency with
     # classic CUL/Apache logs.
-    tz = timezone(config.get('TIMEZONE', 'US/Eastern'))
+    tz = timezone(config.get('ARXIV_BUSINESS_TZ', 'US/Eastern'))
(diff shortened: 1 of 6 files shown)


arxiv/base/logging.py
```diff
@@ -0,0 +1,70 @@
...
+    logger = logging.getLogger(name)
+    config = get_application_config()
+
+    # Set the log level from the Flask app configuration.
+    level = int(config.get('LOGLEVEL', logging.INFO))
+    logger.setLevel(level)
+
+    # Log messages should be in Eastern local time, for consistency with
+    # classic CUL/Apache logs.
+    tz = timezone(config.get('TIMEZONE', 'US/Eastern'))
```
mhl10: Would it make sense to have something like `ARXIV_BUSINESS_TZ` explicitly defined as 'US/Eastern' in config.py?

erickpeirson: Indeed!

erickpeirson:
@mhl10 Looking better after those changes?

PR #49, opened 2018-04-19
ARXIVNG-281 introduced base class and tools for Kinesis consumers

erickpeirson:
(from docs in PR)

> We use Kinesis streams to broker notifications that concern multiple services
in the arXiv system. For example, at publication time, the publication process
generates notifications about new arXiv papers so that services like search
can update themselves.

> This module provides a base class for building agents with Kinesis streams,
:class:`.BaseConsumer`. The objective is to provide all of the
stream-management (e.g. checkpointing) and error-handling boilerplate needed
for any Kinesis consumer, so that we can focus on building the idiosyncratic
functional bits in arXiv services.

> You (the developer) should be able to create a minimal agent by:

> 1. Defining a class that inherits from :class:`.BaseConsumer`
> 2. Defining an instace method on that class with the signature:
   ``def process_record(self, record: dict) -> None:`` that implements
   application-specific processing for each notification.
> 3. Calling :func:`.process_stream` with your consumer class and an application
   config (e.g. from Flask).

### DIFF
--- a/Pipfile
+++ b/Pipfile
@@ -15,6 +15,7 @@ pylint = "*"
 pytz = "*"
 uwsgi = "*"
 pydocstyle = "*"
+"boto3" = "*"
 
 
 [dev-packages]
--- a/Pipfile.lock
+++ b/Pipfile.lock
@@ -1,7 +1,7 @@
 {
     "_meta": {
         "hash": {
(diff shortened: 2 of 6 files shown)


JaimieMurdock:
Thanks for the documentation as well!

PR #53, opened 2018-05-10
Add license definitions to arxiv-base and simple test

mhl10:
This is mostly a conversion of the essentials from `arxiv-lib:arXiv::Config::Licenses`.

### DIFF
--- a/arxiv/license.py
+++ b/arxiv/license.py
@@ -73,7 +73,7 @@ NO_LICENSE_TEXT = 'I do not certify that any of the above licenses apply'
 
 ASSUMED_LICENSE = LICENSES['http://arxiv.org/licenses/assumed-1991-2003/']
 
-# Historical license to updated/current license (old: new)
+# Historical license to updated/current license (old URI: new URI)
 TRANSLATED_LICENSES = {
     'http://creativecommons.org/licenses/by/3.0/':
     'http://creativecommons.org/licenses/by/4.0/',
--- a/arxiv/license.py
+++ b/arxiv/license.py
@@ -6,7 +6,7 @@ LICENSES = {
     'http://arxiv.org/licenses/nonexclusive-distrib/1.0/': {
         'uri': 'http://arxiv.org/licenses/nonexclusive-distrib/1.0/',
         'label': 'arXiv.org perpetual, non-exclusive license to '
-        'distribute this article',
+                 'distribute this article',
         'note': '(Minimal rights required by arXiv.org. Select this '
                 'unless you understand the implications of other '
                 'licenses)',
@@ -34,7 +34,7 @@ LICENSES = {
         'uri': 'http://creativecommons.org/licenses/by-nc-sa/4.0/',
         'order': 7,
         'label': 'Creative Commons Attribution-Noncommercial-ShareAlike '
-        ' license',
+                 'license',
         'is_current': True,
         'is_valid': True,
         'icon_uri': f'{LICENSE_ICON_BASE_URI}/by-nc-sa-4.0.png'},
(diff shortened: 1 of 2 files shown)


arxiv/license.py
```diff
@@ -0,0 +1,114 @@
...
+    'http://creativecommons.org/licenses/by/4.0/',
+    'http://creativecommons.org/licenses/by-nc-sa/3.0/':
+    'http://creativecommons.org/licenses/by-nc-sa/4.0/',
+    'http://creativecommons.org/licenses/publicdomain/':
+    'http://creativecommons.org/publicdomain/zero/1.0/'
+}
+
+CURRENT_LICENSES = {
+    k: v for k, v in LICENSES.items()
+    if all(k in v for k in ('is_current', 'order')) and v['is_current']
```
erickpeirson: This is totally nit-picky, but any reason not to make this more brutish for the sake of clarity? eg.

```if 'is_current' in v and 'order' in v and v['is_current']```

File: arxiv/license.py [hunk 2]
```diff
@@ -0,0 +1,114 @@
+"""arXiv license definitions."""
+
+
+LICENSE_ICON_BASE_URI = '/icons/licenses'
+LICENSES = {
+    'http://arxiv.org/licenses/nonexclusive-distrib/1.0/': {
+        'uri': 'http://arxiv.org/licenses/nonexclusive-distrib/1.0/',
+        'label': 'arXiv.org perpetual, non-exclusive license to '
+        'distribute this article',
```
erickpeirson: Preserve visual indentation?

```
        'label': 'arXiv.org perpetual, non-exclusive license to '
                 'distribute this article',
```

arxiv/license.py
```diff
@@ -0,0 +1,114 @@
...
+    },
+    'http://creativecommons.org/licenses/by/4.0/': {
+        'uri': 'http://creativecommons.org/licenses/by/4.0/',
+        'label': 'Creative Commons Attribution license',
+        'order': 5,
+        'is_current': True,
+        'is_valid': True,
+        'icon_uri': f'{LICENSE_ICON_BASE_URI}/by-4.0.png'},
+
+    'http://creativecommons.org/licenses/by-sa/4.0/': {
```
erickpeirson: Not really a problem, but curious why the URIs are used as keys? Seems like this struct could also be a list of dicts?

erickpeirson: In that case, you could just use list order as de-facto order -- would make it easier to change later, and then you wouldn't need the sort in ``CURRENT_LICENSE_URIS``?

arxiv/tests/test_license.py
```diff
@@ -0,0 +1,21 @@
+"""Tests for arXiv taxonomy module."""
+from unittest import TestCase
+from arxiv.license import LICENSES, CURRENT_LICENSE_URIS
+
+
+class TestLicense(TestCase):
+    """Tests for the arXiv license definitions."""
+
+    def test_current_license_uris(self):
```
erickpeirson: This is basically a regression test? Will it need to be updated when the ordering changes? 

Might consider some other validation checks here, for future changes to the licenses. E.g. assert that all items have uris with truthy str values, assert that all items have descriptions, etc. Thinking about things that should be universally true of licenses in this module, and for which there is a non-negligible probability of bungling in the future.

arxiv/license.py
mhl10: Thanks, good eye.

arxiv/license.py
mhl10: Agree!

arxiv/tests/test_license.py
mhl10: This is indeed a regression test (and port of an old test) and would have to be updated with ordering changes. It probably creates more annoyance than actual value, so I'd be fine with removing it.

I've added some validation checks.

PR #58, opened 2018-06-13
ARXIVNG-913 extended config_url to support a broader set of use-cases

erickpeirson:
Delivers ARXIVNG-913:

> Currently the ``config_url_builder`` context processor (https://github.com/cul-it/arxiv-base/blob/master/arxiv/base/context_processors.py#L11) just gets a parameter from the app config and returns its value. We also need to support cases in which the path must be extended (e.g. to introduce an arXiv ID as a path parameter) or in which request parameters must be added. We can look to `urllib.urlparse` for guidance on how this should work: https://docs.python.org/3/library/urllib.parse.html


arxiv/base/urls.py
```diff
@@ -0,0 +1,73 @@
...
+
+    .. code-block:: python
+
+       from arxiv.base.urls import config_url
+    """
+    target = target.upper()
+    # Look for the URL on the config of the current app (this will *not* be
+    # base); fall back to the base config if not found.
+    try:
+        url: str = current_app.config.get(f'ARXIV_{target}_URL')
```
bdc34: Just to check how this works: the URL paths are stored as string templates in the flask app context and these are used by just running format on them with the params dict.

I'm not sure but it looks like this is separated from the flask blueprint and routes mechanisms. Is it seperated from those or some how integrated?

arxiv/base/urls.py
```diff
@@ -0,0 +1,73 @@
...
+        raise ConfigurationError(f'URL for {target} not set') from e
+
+    # Format with path parameters.
+    try:
+        url = url.format(**path_params)
+    except KeyError as e:
+        raise ValueError('Missing a required parameter: %s' % e) from e
+
+    # Format with request parameters.
+    parts = url_parse(url)
```
bdc34: Please add more of a comment about what is going on here. Why url_parse then url_unparse?

arxiv/base/urls.py
bdc34: An idea for something similar that does use routes would be to define the public routes of other services in base and then redefine then in the actual services:
 ```
In [28]: app = create_web_app()

In [29]: app.url_map
Out[29]: 
Map([<Rule '/test' (HEAD, OPTIONS, GET) -> ui.test_page>,
 <Rule '/404' (HEAD, OPTIONS, GET) -> ui.test_404>,
 <Rule '/403' (HEAD, OPTIONS, GET) -> ui.test_403>,
 <Rule '/401' (HEAD, OPTIONS, GET) -> ui.test_401>,
 <Rule '/405' (HEAD, OPTIONS, GET) -> ui.test_405>,
 <Rule '/413' (HEAD, OPTIONS, GET) -> ui.test_413>,
 <Rule '/400' (HEAD, OPTIONS, GET) -> ui.test_400>,
 <Rule '/500' (HEAD, OPTIONS, GET) -> ui.test_500>,
 <Rule '/static/base/<filename>' (HEAD, OPTIONS, GET) -> base.static>,
 <Rule '/static/<filename>' (HEAD, OPTIONS, GET) -> static>])

In [30]: def pdf(id): print("I'm a placeholder")

In [31]: app.add_url_rule('/pdf/<string:id>', 'pdf',pdf)

In [32]: app.url_map
Out[32]: 
Map([<Rule '/test' (HEAD, OPTIONS, GET) -> ui.test_page>,
 <Rule '/404' (HEAD, OPTIONS, GET) -> ui.test_404>,
 <Rule '/403' (HEAD, OPTIONS, GET) -> ui.test_403>,
 <Rule '/401' (HEAD, OPTIONS, GET) -> ui.test_401>,
 <Rule '/405' (HEAD, OPTIONS, GET) -> ui.test_405>,
 <Rule '/413' (HEAD, OPTIONS, GET) -> ui.test_413>,
 <Rule '/400' (HEAD, OPTIONS, GET) -> ui.test_400>,
 <Rule '/500' (HEAD, OPTIONS, GET) -> ui.test_500>,
 <Rule '/static/base/<filename>' (HEAD, OPTIONS, GET) -> base.static>,
 <Rule '/static/<filename>' (HEAD, OPTIONS, GET) -> static>,
 <Rule '/pdf/<id>' (HEAD, OPTIONS, GET) -> pdf>])```

erickpeirson: That's an excellent idea. Will take a stab and update the PR

bdc34: Great!

erickpeirson:
@mhl10 @bdc34 Finally got back around to this. I think this gets us what we're looking for. The main limitation of using ``url_for`` is that it expects to be working with relative URLs. So I left the ``config_url`` function for external links, implemented the suggested ``add_url_rule`` approach, and cleaned up the config file a bit.

Somewhat tangential, but I also made a slight tweak to the package structure, so that we can use the ``arxiv`` namespace in other packages. This shouldn't have any impact on projects that are using this package already.

PR #63, opened 2018-07-06
ARXIVNG-962 adds arxiv.forms, which provides tools for CSRF protection

erickpeirson:
To protect a form, inherit from`arxiv.forms.csrf.CSRFForm`. When the form is handled,
this class expects that:

- There is an active [Flask request](http://flask.pocoo.org/docs/1.0/reqcontext/).
- The object ``flask.request.session`` exists, and has a ``str`` attribute
  ``nonce``, such as you might find [here](https://cul-it.github.io/arxiv-accounts/arxiv.users/arxiv.users.domain.html#arxiv.users.domain.Session.nonce). See [`arxiv.users.domain.Session`](https://cul-it.github.io/arxiv-accounts/arxiv.users/arxiv.users.domain.html#arxiv.users.domain.Session) and
  [`arxiv.users.auth.Auth`](https://cul-it.github.io/arxiv-accounts/arxiv.users/arxiv.users.auth.html#arxiv.users.auth.Auth).
- A parameter called ``CSRF_SECRET`` is set in the application configuration.

Here's an example:

```python
   from typing import Tuple
   from werkzeug import MultiDict
   from wtforms import StringField
   from arxiv import status
   from arxiv.forms import csrf

   ResponseData = Tuple[dict, int, dict]


   class ProtectedForm(csrf.CSRFForm):
       """A CSRF-protected form."""

       something_sensitive = StringField('Something sensitive')


   def some_controller(method: str, form_data: MultiDict) -> ResponseData:
       """Handle a form-based view."""
       headers = {}
       if method == 'POST':
           form = ProtectedForm(form_data)
           data = {'form': form}
           if not form.validate():    # Checks the CSRF token, too!
               return data, status.HTTP_400_BAD_REQUEST, headers

           # do something sensitive
           return data, status.HTTP_201_CREATED, headers

       form = ProtectedForm()
       return {'form': form}, status.HTTP_200_OK, headers
```

And in your template:

```html
   {% extends "base/base.html" %}
   {% block content %}
   <form method="POST" action="{{ url_for('ui.some_view') }}">
     {{ form.csrf_token }}
     {{ form.something_sensitive }}
     <input type="submit" class="button is-link is-small"></input>
   </form>
   {% endblock content %}
```

For more information about CSRF mechanisms in WTForms, see http://wtforms.readthedocs.io/en/latest/csrf.html#module-wtforms.csrf

@DavidLFielding @eawoods Tagging you guys as this was driven by submission-related concerns. 
@mhl10 Looking for another set of eyes from the Flask side, and also since you're looking at https://github.com/cul-it/arxiv-accounts/pull/12


arxiv/forms/csrf.py
```diff
@@ -0,0 +1,153 @@
...
+
+    @staticmethod
+    def _split(csrf_token: str) -> Tuple[str, str]:
+        digest, expires = csrf_token.split('::', 1)
+        return digest, expires
+
+    def generate_csrf_token(self, field: 'CSRFForm') -> str:
+        """Generate a new CSRF token using the CSRF secret and context."""
+        expires = self._new_expiry()
+        nonce = self.csrf_context['nonce']
```
erickpeirson: Consider using legacy session ID instead (since we may not have NG sessions in place when submission goes out)

PR #64, opened 2018-07-16
ARXIVNG-1005 reimplementation of URL mapping for external endpoints

erickpeirson:
This addresses #66, and implements mapping for external URLs in a way that is more flexible and also more in line with the Flask API. The basic idea is that instead of directly manipulating the application's URL map (which can't be changed later on), we use a hook provided by Flask to try to handle endpoints that cannot be resolved by [``url_for``](http://flask.pocoo.org/docs/1.0/api/#flask.url_for). Under the hood, it uses Werkzeug's routing tools (e.g. ``werkzeug.routing.Map``, which is what Flask uses for ``Flask.url_map``) so that we are using consistent path formatting patterns. 

Another advantage of this approach is that there is a clear way to use map to external URLs; endpoints need not be at the same hostname. This obviates the need for the separate ``config_url`` function, so I removed it. This way there is only one way to generate URLs: Flask's built-in ``url_for`` function. In order to maintain support for env-defined URLs, I added a few lines to ``arxiv.base.config`` to look for and incorporate config_url-style endpoint definitions.

Since this removes ``config_url``, I bumped the version to 0.9.1.


arxiv/base/urls.py
```diff
@@ -1,80 +1,111 @@
-"""Helpers for working with URLs in arXiv Flask applications."""
+"""
+Helpers for working with URLs in arXiv Flask applications.
```
erickpeirson: How-to start here

arxiv/base/__init__.py
```diff
@@ -63,15 +57,10 @@ def init_app(self, app: Flask) -> None:
...
             app.errorhandler(error)(handler)
 
-        # Register URLs for other services. This allows us to use flask's
-        # ``url_for()`` function to generate URLs for services that are
-        # deployed at the same hostname.
-        for name, pattern in config.ARXIV_URLS:
-            app.add_url_rule(pattern, name, placeholder)
+        # Attach the external URL handler as a fallback for failed calls to
+        # url_for().
+        app.url_build_error_handlers.append(urls.external_url_handler)
```
erickpeirson: Here's where we hook into Flask

arxiv/base/config.py
```diff
@@ -1,46 +1,51 @@
...
+    ("account", "/user", "arxiv.org"),
+    ("login", "/user/login", "arxiv.org"),
+    ("logout", "/user/logout", "arxiv.org"),
+    ("home", "/", "arxiv.org"),
+    ("pdf", "/pdf/<arxiv:paper_id>", "arxiv.org"),
+    ("twitter", "/arxiv", "twitter.com"),
+    ("blog", "/arxiv", "blogs.cornell.edu"),
+    ("wiki", "/display/arxivpub/arXiv+Public+Wiki", "confluence.cornell.edu"),
+    ("library", "/", "library.cornell.edu"),
+    ("acknowledgment", "/x/ALlRF", "confluence.cornell.edu")
```
bdc34: Since this hasn't been merged yet, can you add the following URLs?

@blueprint.route("help_identifier", "/help/arxiv_identifier", "arxiv.org")
@blueprint.route("help_mathjax", "/help/mathjax/", "arxiv.org")
@blueprint.route("help_trackback", "/help/trackback/","arxiv.org")
@blueprint.route("help_social_bookmarking", "/help/social_bookmarking","arxiv.org")

# Not sure how to do static
@blueprint.route("arxiv_css", "/css/arXiv.css?v=20170424","static.arxiv.org")

# Not sure how to do the favicon
@blueprint.route("favicon", /favicon.ico, "arxiv.org")

# deprecated? 
@blueprint.route("", "/find","arxiv.org")

# anti-robot honnypot 
@blueprint.route("ignore_me", "/IgnoreMe","arxiv.org")

arxiv/base/config.py
```diff
@@ -1,46 +1,51 @@
...
-    ("help", "/help"),
-    ("contact", "/help/contact"),
-    ("search_box", "/search"),
-    ("search_advanced", "/search/advanced"),
-    ("account", "/user"),
-    ("login", "/user/login"),
-    ("logout", "/user/logout"),
-    ("home", "/"),
-    ("pdf", "/pdf/<arxiv:paper_id>"),
+URLS = [
```
bdc34: It seems like we'll need to have something like:
BASE_SERVER='arxiv.org'
because in many cases it will need to be configured:
BASE_SERVER='beta.arxiv.org'

erickpeirson:
Added the help routes and honeypot, and added the ``BASE_SERVER`` parameter. 

For the shared static files, I created https://culibrary.atlassian.net/browse/ARXIVNG-1015. The short story is that this will provide reverse URL resolution for shared static files in S3 without having to manually configure them here. Tapping @mhl10 on that, as he set this up for search.

PR #66, opened 2018-07-20
ARXIVNG-1005 update Bulma to version 0.7.1, update global styles

eawoods:
In this PR:
- Bulma 0.7.1 has some new variables and controls, added with new version. Also has better support for semibold font weight, which looks terrific (can be invoked with has-weight-semibold if needed).
- I took the opportunity to move Bulma into its own directory to prevent us from modifying it, thus having overwrites occur when we upgrade over time.
- Opportunistic moves of some code from arxiv-search were taken (listed as #TODO in arxiv-search sass).
- Added better support for message styling the way our default design looks (via Bulma's variables).
- *NOTE: Changed default base text size* from 14px to 16px to support accessible design (it's supposed to be that big). I am intending to do a thorough cross-repo review on Monday morning to make sure search layout didn't get broken in bad ways by this subtle but important change, since all text styles are defined relative to the base.
- H1 size and default width changed slightly so that h2 is no longer visually bigger than h1.

screen shot 2018-07-20 at 4 29 02 pm

Edit 7-23-18: added search result preview to confirm font size changes (this would be achieved by overriding the current breathe-horizontal to be 60% of current width, and including the mini form at the top in the breathe-horizontal margin space, requires a small adjustment to search templates).
screen shot 2018-07-23 at 10 33 53 am


eawoods:
Oh yeah, and I also created styleguide.html as a template and bound it to the /test route. A nice way for me to create a living style guide, provide a platform for testing display of elements, and document display conventions.

arxiv/base/static/css/arxivstyle.css
```diff
@@ -2457,7 +2573,7 @@ a.box:active {
   .help.is-dark {
     color: #2d2d2d; }
   .help.is-primary {
-    color: #711111; }
+    color: #b31b1b; }
```
mhl10: This looks more like arXiv red :-)

eawoods: It's official Cornell red :)

arxiv/base/routes.py
```diff
@@ -20,7 +20,7 @@
 @blueprint.route('/test', methods=['GET'])
```
mhl10: This seems fine, though could be something more specific (/styleguide)?

arxiv/base/static/sass/bulma/README.md
```diff
@@ -0,0 +1,106 @@
+# [Bulma](https://bulma.io)
```
erickpeirson: Should this be in here?

arxiv/base/routes.py
eawoods: Good thought, I'll change it! I had repurposed an old route, but you're right that it isn't descriptive of the current function.

arxiv/base/static/sass/bulma/README.md
eawoods: I'm not sure -- advise me on GitHub manners?  Is it better to have the entirety of the bulma package intact, as authored, or remove the readme and changelog so that the files don't accidentally get picked up by any of our documentation?
I'm good either way. Suppose that separating the Bulma files into their own directory is enough separation to indicate their collective function as a library.

PR #70, opened 2018-08-10
ARXIVNG-1058 implemented flash messages in base templates

erickpeirson:
This implements consistent flash messages in base templates. I introduced the module ``arxiv.base.messages`` to ensure consistent implementation. This extends the [Flask built-in flash messages](http://flask.pocoo.org/docs/1.0/patterns/flashing/) with structured messages. 

I also added a pure-JavaScript dismiss feature for the dismissible notifications.

This is up to date with https://github.com/cul-it/arxiv-base/pull/69, so the changed files will reflect changes in that PR until it is merged.

image

To see this in action, start the dev server and visit http://127.0.0.1:5000/styleguide. You won't see anything when you first land, but the messages should appear when you refresh the page.


arxiv/base/messages.py
```diff
@@ -0,0 +1,157 @@
...
+    _flash_with(WARNING, message, title, dismissable, safe)
+
+
+def flash_failure(message: str, title: Optional[str] = None,
+                  dismissable: bool = True, safe: bool = False) -> None:
+    """
+    Flash a failure message to the user.
+
+    Failures are an unavoidable part of life. We should be honest about our
+    mistakes, and find constructive ways to move forward with our lives.
```
eawoods: ha, love the attitude. Even though this is a comment, it's a set of directions for using the module: perhaps a short statement about failure messages containing hints for success? (Accessibility requirement, possibly worth mentioning)

erickpeirson:
Looks like the failing build is due to connection issues with dockerhub -- all tests passing

arxiv/base/messages.py
```diff
@@ -0,0 +1,157 @@
...
+    """
+    _flash_with(FAILURE, message, title, dismissable, safe)
+
+
+def flash_success(message: str, title: Optional[str] = None,
+                  dismissable: bool = True, safe: bool = False) -> None:
+    """
+    Flash a success message to the user.
+
+    It's nice to know when you've done something right for a chance.
```
eawoods: typo:  change

arxiv/base/static/js/notification.js
```diff
@@ -0,0 +1,27 @@
+/**
+  * Implements dismissable notifications.
+  *
+  * To use, insert an element (e.g. a button) with the class
+  * `notification-dismiss` that is the direct child of the element that you
```
eawoods: Does the `notification-dismiss` class exist solely to drive the JavaScript for dismissal? If so - does it cause collisions with any other variables to use the Bulma default class `delete` instead? (I recognize it's a common word and may not be usable as such.)

arxiv/base/static/sass/_custom.sass
```diff
@@ -212,3 +212,21 @@ tr.is-warning
 
 .pagination-list li + li // fixes override from content li + li
   margin: 0
+
+/* flash notifications */
+
+.notifications
+  margin-top: 5px
```
eawoods: whenever possible, specify all dimensions relatively (for example .3em)

arxiv/base/static/sass/_custom.sass
```diff
@@ -212,3 +212,21 @@ tr.is-warning
...
+
+.notification
+  padding: 0.25em 0.5em 0.25em 0.5em
```
eawoods: notification padding is a Bulma variable that can be set in arxivstyle.sass - use `$notification-padding`

arxiv/base/static/sass/_custom.sass
```diff
@@ -212,3 +212,21 @@ tr.is-warning
...
+  width: 70%
+  margin-left: auto
+  margin-right: auto
+  margin-bottom: 5px !important
+
+.notification .notification-message
```
eawoods: possibly can use `.is-size-7` instead to keep number of classes down

arxiv/base/static/sass/_custom.sass
```diff
@@ -212,3 +212,21 @@ tr.is-warning
...
+  font-size: 0.8em
+
+.notification .notification-title
+  font-weight: 600
```
eawoods: definitely can use `.has-text-weight-bold` or specify heading level as appropriate (better semantic HTML for a11y)

arxiv/base/templates/base/header.html
```diff
@@ -13,6 +13,24 @@ <h1 class="level-item"><a href="{{ url_for('home') }}" aria-label="arxiv-logo">a
...
-<!--  # TODO: reintroduce this once we have access to the user's session.
-  <div class="user-tools box is-pulled-right" role="navigation" aria-label="User menu"><a href="{{ url_for('login') }}">Login</a> | <a href="{{ url_for('account') }}">My Account</a> | <a href="{{ url_for('logout') }}">Logout</a></div>
--->
+<div class="is-pulled-right">
+  <div class="user-tools box level-item" role="navigation" aria-label="User menu">
+    {% if request.session and request.session.user %}Logged in as <strong>{{ request.session.user.username }}</strong> | <a href="{{ url_for('account') }}">My Account</a> | <a href="{{ url_for('logout') }}">Logout</a>{% else %}<a href="{{ url_for('login') }}">Login</a>{% endif %}
+  </div>
+</div>
+
+<div class="container is-fluid notifications">
```
eawoods: I would suggest having the notification width be set at this level -- to avoid possibly making all notifications in other contexts an undesired width.
Also: add `role=alert` to this div to have screen readers interrupt flow of document to bring flash messages first.

arxiv/base/templates/base/header.html
```diff
@@ -13,6 +13,24 @@ <h1 class="level-item"><a href="{{ url_for('home') }}" aria-label="arxiv-logo">a
...
+    {% with messages = get_flashed_messages(with_categories=true) %}
+      {% if messages %}
+        {% for category, message in messages %}
+          <div class="notification is-{{ category }}">
```
eawoods: perfect

arxiv/base/templates/base/header.html
```diff
@@ -13,6 +13,24 @@ <h1 class="level-item"><a href="{{ url_for('home') }}" aria-label="arxiv-logo">a
...
+            {% if message.dismissable %}<button class="delete notification-dismiss"></button>{% endif %}
+            {% if message.title %}<div class="notification-title">{{ message.title }}</div>{% endif %}
```
eawoods: Yeah, this should probably be at some heading level for a11y. I'll review onscreen to figure out which one it should be to avoid skipping heading levels.

arxiv/base/templates/base/header.html
```diff
@@ -13,6 +13,24 @@ <h1 class="level-item"><a href="{{ url_for('home') }}" aria-label="arxiv-logo">a
...
+            <p class="notification-message">{{ message.message }}</p>
```
eawoods: Suggestion: could use .is-text-7 or set all p children of notification to the same level without defining a new class.

arxiv/base/static/js/notification.js
erickpeirson: Yeah, ``delete`` is pretty broad. I preferred ``notification-dismiss`` for semantic clarity

arxiv/base/templates/base/header.html
erickpeirson: I'm having trouble getting this to work at the container level. If I use ``!important`` when setting width at that level, I am able to get the desired width. But because of the way centering works with the container it gets pulled offcenter to the left. Tempted to leave this for now, and you may have better luck?

erickpeirson:
@eawoods Ok! I was able to address everything (I think) except for the width issue (comment above). I struggled a little bit with how the container element is supposed to work, and what Bulma means when it says that content is centered in a container (lies, I say -- or at least a flexible interpretation of the truth).

PR #72, opened 2018-08-20
ARXIVNG-1088 notification styling

eawoods:
I could mess around with base for a loooooong time and never be completely satisfied.
Probably best to stick with the task at hand.

Notifications are softer:

screen shot 2018-08-20 at 11 17 43 am


arxiv/base/factory.py
```diff
@@ -13,4 +13,5 @@ def create_web_app() -> Flask:
     app.config.from_pyfile(os.path.join(basedir, 'config.py'))
     Base(app)    # Gives us access to the base UI templates and resources.
     app.register_blueprint(routes.blueprint)
+    print(app.url_map)
```
erickpeirson: Looks like this is left over from debuging :-)

eawoods: whoops! Fixed.

PR #79, opened 2018-12-06
Moves jinja2 template macros for abs page parts from arxiv-browse to arxiv-base

bdc34:
Moves jinja2 template macros for abs page parts from arxiv-browse to arxiv-base.


arxiv/browse/templates/abs_macros.html
```diff
@@ -0,0 +1,153 @@
...
+    {% if abs_meta.proxy != None and abs_meta.proxy|wordcount > 1 %}&nbsp;[via {{ abs_meta.proxy.split()[0]|tex_to_utf|upper }} proxy]{% endif %}
+    <br/>
+    {%- for version_entry in abs_meta.version_history -%}
+      {{ generate_version_entry(abs_meta, version_entry, abs_meta.version) }}
+    {%- endfor -%}
+  </div>  
+{%- endmacro -%}
+
+
+{%- macro abstract(abs_meta) %}
```
erickpeirson: From ARXIVNG-1389: 
> This should be a macro in here: https://github.com/cul-it/arxiv-base/blob/master/arxiv/base/templates/base/macros.html 
> Each element (title, abstract, etc) should passed as a separate parameter to the macro.

Per Slack #arxiv-browse on Thursday,  November 29:
> what we need here is to move over the abs preview as a template macro that takes native Python objects as arguments -- no domain classes, etc, as those are specific to the domain of the browse service.

arxiv/browse/util/author_affil.py
```diff
@@ -0,0 +1,363 @@
+"""Parse Authors lines to extract author and affiliation data."""
```
erickpeirson: Per Slack discussion 26 November and comments in ARXIVNG-1097: author name parsing should be added to ``arxiv.util``. ``arxiv.util.author_affil`` would be ok.

arxiv/browse/util/id_patterns.py
```diff
@@ -0,0 +1,329 @@
+r"""Patterns and functions to detect arXiv ids and Urls in text.
```
erickpeirson: Per Slack discussion in #arxiv-browse Friday, November 30th:
> - `browse.util.id_patterns` -> `arxiv.urlize`

arxiv/browse/util/search_authors.py
```diff
@@ -0,0 +1,166 @@
+"""Produces search service query strings for authors."""
```
erickpeirson: If this is about url generation (looks like it, in essence), it should be in ``arxiv.urlize`` with the arXiv ID and DOI link generation code.

arxiv/browse/util/tex2utf.py
```diff
@@ -0,0 +1,186 @@
+"""Convert between TeX escapes and UTF8."""
```
erickpeirson: Per Slack on Friday, 30 November:
> - `browse.services.util.tex2utf` -> `arxiv.util`

``arxiv.util.tex2utf`` would be fine

erickpeirson:
Closing per #81

PR #86, opened 2019-01-14
ARXIVNG-971 added ISO8601JSONEncoder and ISO8601JSONDecoder

erickpeirson:
Adds ``arxiv.util.serialize.ISO8601JSONEncoder`` and ``.ISO8601JSONDecoder``, and convenience functions ``.loads`` and ``.dumps``.  For background on JSON encoders/decoders, see https://docs.python.org/3/library/json.html#encoders-and-decoders.

### DIFF
--- a/arxiv/util/tests/test_serialize.py
+++ b/arxiv/util/tests/test_serialize.py
@@ -1,0 +1,128 @@
+"""Tests for :mod:`.serialize`."""
+
+from unittest import TestCase, mock
+from datetime import datetime, date
+from pytz import UTC, timezone
+from .. import serialize
+
+ET = timezone('US/Eastern')
+
+
+class TestISO8601JSONEncoder(TestCase):
+    """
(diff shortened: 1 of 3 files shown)


arxiv/util/tests/test_serialize.py
```diff
@@ -0,0 +1,128 @@
...
+            serialize.dumps([{"date": datetime(year=1993, month=11, day=3)}]),
+            '[{"date": "1993-11-03T00:00:00"}]'
+        )
+        self.assertEqual(
+            serialize.dumps([{"date": [datetime(year=1993, month=11, day=3)]}]),
+            '[{"date": ["1993-11-03T00:00:00"]}]'
+        )
+
+    def test_encode_with_datetime_tz(self):
+        """Encode JSON with a TZ-aware :class:`.datetime` object."""
```
mhl10: Thanks for including tests with `tzinfo` set; it was the first thing I was looking for!

erickpeirson:
@mhl10  Does that param do what you want?

PR #87, opened 2019-01-18
ARXIVNG-250 Linkification in abs view should not choke on parentheses

erickpeirson:
I refactored ``arxiv.base.urls.links`` to use the [``bleach``](https://bleach.readthedocs.io/en/latest/) package from Mozilla. It is a little more user-friendly than maintaining our own core code for matching/replacing identifiers and/or URLs. I combined the regex that was already there for DOIs, arXiv IDs, and URL cases not already handled by bleach with the core bleach URL regex. I then wrote some [callbacks](https://bleach.readthedocs.io/en/latest/linkify.html?highlight=callbacks#callbacks-for-adjusting-attributes-callbacks) to alter link text, add CSS classes (new), add ``rel`` attribute, generate URLs for arXiv IDs and callback URLs for DOIs, etc.

The tests are basically the same, I just updated them to reflect the classes, rel, and other attributes. But the matching/replacing behavior is just as before, with the improvement that (ARXIVNG-250) parentheses are now supported in URLs.

Whew!

### DIFF
--- a/arxiv/base/urls/links.py
+++ b/arxiv/base/urls/links.py
@@ -135,7 +135,8 @@ def _add_rel_external(attrs: Mapping, new: bool = False) -> Mapping:
         attrs[(None, 'rel')] = 'external'
         attrs['_text'] = f'this {o.scheme} URL'    # Replaces the link text.
         _extend_class_attr(attrs, 'link-external')
-    elif (None, 'data-arxiv') not in attrs and (None, 'data-doi') not in attrs:
+    elif (None, 'data-arxiv-id') not in attrs \
+            and (None, 'data-doi') not in attrs:
         attrs['_text'] = f'this {o.scheme} URL'    # Replaces the link text.
         _extend_class_attr(attrs, 'link-internal')
     return attrs
@@ -170,7 +171,7 @@ def _handle_arxiv_url(attrs: Mapping, new: bool = False) -> Mapping:
     if ARXIV_ID.match(target):
         attrs[(None, 'href')] = arxiv_id_to_url(target)
         attrs['_text'] = f'{prefix}{target}'
-        attrs[(None, 'data-arxiv')] = target     # Add arxiv="<arxiv id>"
+        attrs[(None, 'data-arxiv-id')] = target     # Add arxiv="<arxiv id>"
     return attrs
 
 
--- a/arxiv/base/urls/links.py
+++ b/arxiv/base/urls/links.py
@@ -61,9 +61,9 @@ def _identity(x: str) -> str:
     return x
 
 
+DOI = re.compile(r'(?P<doi>10.\d{4,9}/[-._;()/:A-Z0-9]+)', re.I)
 doi_patterns = [
-    Matchable(['10.1145/0001234.1234567'],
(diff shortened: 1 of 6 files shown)


arxiv/identifier/__init__.py
```diff
@@ -25,23 +25,24 @@
 )
 
 OLD_STYLE_WITH_ARCHIVE = re.compile(
-    r'%s(?P<arxiv_id>(%s)\/\d{2}[01]\d{4}(v\d*)?)' % (_prefix, _archive),
```
mhl10: If the named groups (like `P<arxiv_id>`) are being removed from the regexes, then anything that uses them (see `parse_arxiv_id` below) should be updated to not use them. Personally, I find named groups more readable.

erickpeirson:
Good point. We need tests for arxiv.identifier.

I took them out because I wanted to combine them into one pattern, but
duplicate keys aren't allowed. I'll be less lazy and fine a better way :-)

arxiv/base/urls/links.py
```diff
@@ -74,242 +60,210 @@ def _identity(x: str) -> str:
...
-def _to_tags(targets: Tuple[str, List[str], Substituter, Izer],
-             bad_patterns: List[Pattern], text: str)-> str:
-    """Split text to tokens, do _transform_token for each, return results."""
-    transform_token = partial(_transform_token, targets, bad_patterns)
+        target = href
+        prefix = ''
+    if DOI.match(target):
+        attrs[(None, 'href')] = clickthrough_url_for_doi(target)
+        attrs['_text'] = f'{prefix}{target}'
+        attrs[(None, 'doi')] = target     # Add arxiv="<arxiv id>"
```
mhl10: `data-doi` per earlier comment

arxiv/identifier/__init__.py
erickpeirson: Reverted this; added a function in ``arxiv.base.urls`` to remove the group names so that I can re-use them without breaking ``arxiv.identifier``.

erickpeirson: Oh, and now we have some tests in ``arxiv.identifier``!

arxiv/base/urls/links.py
```diff
@@ -74,242 +60,210 @@ def _identity(x: str) -> str:
...
+        target = href.split(':', 1)[1]
+        prefix = 'arXiv:'
     else:
-        back = match.string[match.end():]
+        target = href
+        prefix = ''
+    if ARXIV_ID.match(target):
+        attrs[(None, 'href')] = arxiv_id_to_url(target)
+        attrs['_text'] = f'{prefix}{target}'
+        attrs[(None, 'arxiv')] = target     # Add arxiv="<arxiv id>"
```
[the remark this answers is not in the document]
erickpeirson: Yeah, good point. Changed

arxiv/base/urls/links.py
erickpeirson: Changed

erickpeirson:
@mhl10 @eawoods Ok, I think this is ready for a second look.

arxiv/identifier/__init__.py
mhl10: Thanks!

PR #98, opened 2019-02-15
ARXIVNG-1669 new fourohfour app, ARXIVNG-1473 conditional deployment to k8s

erickpeirson:
This adds a 404 app that can be used as a default backend in Kubernetes (i.e. when no service matches the request). The app is in ``fourohfour/``, and includes a Dockerfile. It has a ``healthz`` endpoint that returns 200 so that we can make sure it's up.

This also adds a script that deploys to development and staging, similar to what you saw in https://github.com/arXiv/arxiv-docs/pull/45.


erickpeirson:
We can do a squash-merge on this

mhl10:
Just so we have a model example:
https://github.com/arXiv/arxiv-base/blob/d60e1a18037207801487bad5044e9467f56f1d41/arxiv/base/routes.py#L59

--> `" respectively.\n  This has an \"abstract\" linefeed.""",`

erickpeirson:
> Has the abstract line feed macro changed? In our existing abstracts, a double \n translates to a paragraph break in the HTML.
> If this abstract is suppose to be a test case representing existing abstracts, a single \n should not trigger a break, and a \n\n should.

No change that I'm aware of. The final version of that line was:

```python
 " respectively.\n\nThis has some linefeeds""",
```

mhl10:
> Has the abstract line feed macro changed? In our existing abstracts, a double \n translates to a paragraph break in the HTML.
> If this abstract is suppose to be a test case representing existing abstracts, a single \n should not trigger a break, and a \n\n should.

I transferred the abstract line feed macro to base. There need to be leading spaces on the new line for the macro to insert the break; see my earlier suggestion above.

erickpeirson:
> > Has the abstract line feed macro changed? In our existing abstracts, a double \n translates to a paragraph break in the HTML.
> > If this abstract is suppose to be a test case representing existing abstracts, a single \n should not trigger a break, and a \n\n should.
> 
> I transferred the abstract line feed macro to base. There need to be leading spaces on the new line for the macro to insert the break; see my earlier suggestion above.

Ah missed that, sorry

bdc34:
That doesn't seem to be what is in develop. 
Please check for the correct double \n test string here in  [arxiv/base/routes.py on develop](https://github.com/arXiv/arxiv-base/blob/8578c489c3e5ca9259f0d3ab859b44723f77b860/arxiv/base/routes.py#L59)

mhl10:
Does the test string example not reflect a case where the filter should be inserting a `<br />`?

https://github.com/arXiv/arxiv-base/blob/develop/arxiv/base/filters.py#L25

bdc34:
Ah, I thought \n\n created the paragraph break. Sorry about that.

mhl10:
No worries, I was confused about it too until I had a closer look.

PR #99, opened 2019-02-18
ARXIVNG-1376 a simple sendmail-based method for sending e-mail 

erickpeirson:
That's pretty much it. Looking at arXiv-Lib, that's pretty much the core of what we're doing (without wonderful HTML). On top of this we can add convenience functions for rendering templates...but basically all you need is:

```python
from flask import render_template
from arxiv import mail

mail.send("foo@user.edu", "Some arXiv for you", 
          render_template("my/template.txt", **context), 
          html_body=render_template("my/template.html", **context))
```

### DIFF
--- a/Pipfile.lock
+++ b/Pipfile.lock
@@ -24,18 +24,18 @@
         },
         "boto3": {
             "hashes": [
-                "sha256:63cd957ba663f5c10ff48ed904575eaa701314f79f18dbc59bd050311cd5f809",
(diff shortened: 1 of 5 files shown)


erickpeirson:
@mhl10 I didn't dig too deeply into our current SMTP settings (aside from noting the host). Would be good to hear what else we'll need to run this on classic. Going to merge for now, but just want to make sure that you see it.

erickpeirson:
Going to merge despite mypy errors; looks like there are several unrelated that need to get cleaned up

PR #101, opened 2019-02-19
ARXIVNG-1673 Jinja templates for email generation

eawoods:
Step 2, templates.

I probably forgot some detail in all of this, but have been staring at this all morning and am ready for some fresh thinking.

I assume that the glue to attach this to the sending function will contain the system messages we want to send and map them to variables the templates can use, and then take the output of the templates into the params that the send function wants. Yes?

Block regions are documented in the README to keep track of it all, since the templates are a bit verbose for figuring out what to do.


arxiv/mail/templates/README.md
```diff
@@ -0,0 +1,66 @@
+# arXiv Email Templates
```
erickpeirson: As someone who doesn't write a lot of HTML e-mails by hand, I find this readme really helpful

arxiv/mail/templates/head.html
```diff
@@ -0,0 +1,81 @@
...
+<!--[if mso]>
+    <style>
+        * {
+            font-family: sans-serif !important;
+        }
+    </style>
+<![endif]-->
+
+<!--[if !mso]>
+<link href='https://fonts.googleapis.com/css?family=Roboto:400,700' rel='stylesheet' type='text/css'>
```
erickpeirson: I'm a little unsure about including refs to third-party resources in our e-mails, since it could be seen as an opportunity for tracking

erickpeirson:
Oh, and the failing `mypy` is unrelated, so don't let that block

arxiv/mail/templates/README.md
eawoods: Oh good! I was hoping to leave a good set of references, because HTML email is so specialized.

arxiv/mail/templates/head.html
eawoods: I'm okay with leaving it out if you don't feel great about it. Many companies and templates do include webfonts, but many also don't. If we don't: we will be at the whims (a bit) of system font defaults. However, if we do: some email clients still don't support webfonts anyway (Gmail started supporting them just this year and that's almost 30% of average viewerships), so I have some decent fallbacks in place.

Basically not a big deal, I'm happy to go with your call on it.

bdc34: I'm against facilitating tracking of arXiv users by third parties.

erickpeirson: > I'm okay with leaving it out if you don't feel great about it. Many companies and templates do include webfonts, but many also don't. If we don't: we will be at the whims (a bit) of system font defaults. However, if we do: some email clients still don't support webfonts anyway (Gmail started supporting them just this year and that's almost 30% of average viewerships), so I have some decent fallbacks in place.
> 
> Basically not a big deal, I'm happy to go with your call on it.
> (I mean yes, the typographer in me sobs about using Arial, but it's all good.)

It's not so much about including fonts as it is about loading the fonts from someone else. So if that font is something that we want to use, we could consider serving it from our own CDN.

eawoods: We certainly could host the fonts ourselves, and this is an issue that should be raised outside of the context of just email.

Note that we are already using third party APIs/CDNs on our web pages (not email, true, but odd that this hasn't come up before):
https://github.com/arXiv/arxiv-base/blob/d68bca8329a6697e34ea508cee76262477b7aa76/arxiv/base/static/sass/arxivstyle.sass#L1
https://github.com/arXiv/arxiv-base/blob/d68bca8329a6697e34ea508cee76262477b7aa76/arxiv/base/templates/base/head.html#L15

eawoods:
Updated: removed call to Google Fonts cdn. I left Roboto in the font stack because some devices (Android) have it as a standard font, and I added LG in the hopes that at least SOME clients will have a more on-brand experience (MacOS, mostly).

However, the font hosting issue needs to be visited in a separate issue because NG's style depends on both Open Sans and FontAwesome. According to their licenses they can both be hosted locally, it would be a task that needs doing.

arxiv/mail/templates/head.html
erickpeirson: You're absolutely right about having done this before, and we should look at hosting these ourselves as a task in itself. Thanks to @bdc34 for helping to keep us on the straight and narrow

PR #107, opened 2019-03-14
ARXIVNG-1970 Provide boilerplate for HTTP integrations in arXiv base

erickpeirson:
We do a lot of HTTP-based service integrations. We want to keep these as evolvable as possible, but there is quite a bit of boilerplate that simply won't change. h/t to @JaimieMurdock who (a while back) suggested moving the exceptions to a shared location.

See docs in code for more.


erickpeirson:
The most interesting part of this PR is here: https://github.com/arXiv/arxiv-base/pull/107/files#diff-ddd9f12ba693d2a35d65dbfeffb14285R45

erickpeirson:
This part is also kind of interesting: https://github.com/arXiv/arxiv-base/pull/107/files#diff-62de0599a9c0a8663514be091cad9c49R6

erickpeirson:
Added all you folks since this is kind of a major cross-cutting thing

arxiv/integration/api/service.py
```diff
@@ -0,0 +1,261 @@
...
+    def _path(self, path: str, query: dict = {}) -> str:
+        """Generate a full path for a request."""
+        o = urlparse(self._endpoint)
+        path = path.lstrip('/')
+        return urlunparse((  # type: ignore
+            o.scheme, o.netloc, f"{o.path}{path}",
+            None, urlencode(query), None
+        ))
+
+    def _check_status(self, resp: requests.Response,
```
bdc34: The HTTP status code handling could be incorporated in a more composable way. It seems that it could be a function such as http_status_to_raised_exception(status:int)->None. Then it could be used anywhere.

arxiv/integration/api/service.py
```diff
@@ -0,0 +1,261 @@
...
+
+    Provides binding to the application context, so that connections can be
+    re-used.
+
+    To implement an HTTP API service integration, define a class that inherits
+    from :class:`.HTTPIntegration`. At minimum, it should include an inner
+    class called ``Meta`` with the attribute ``service_name``; this is used to
+    identify the configuration parameters for the service. For example:
+
+    .. code-block:: python
```
bdc34: The COOL_BEANS_ENDPOINT mapping from a config to a class/instance seems circuitous and unexpected.

It seems like this fixes the implementation of the service at each place that imports and uses it. 

What is the advantage of this over how jinja adds a dict entry to the app? Such as in factory.py create_web_app() doing something like app['CoolIntegration'] = CoolIntegrationImplXyz( ... )

arxiv/integration/api/service.py
```diff
@@ -0,0 +1,261 @@
...
+
+        """
+        response = self.request(method, path, token, expected_code, **kwargs)
+        try:
+            return response.json(), response.status_code, response.headers
+        except json.decoder.JSONDecodeError as e:
+            raise BadResponse(f'Could not decode', response) from e
+
+    def get_status(self) -> dict:
+        """Get the status of the file management service."""
```
bdc34: This probably doesn't need "file management"

arxiv/integration/api/status.py
```diff
@@ -0,0 +1,47 @@
+"""Contants for HTTP status codes. Merely to increase code readability."""
+
+HTTP_100_CONTINUE = 100
```
bdc34: Can we use https://docs.python.org/3/library/http.html#http.HTTPStatus instead?

arxiv/integration/meta.py
```diff
@@ -0,0 +1,146 @@
...
+               return get_session()
+           elif 'filemanager' not in g:
+               g.filemanager = get_session()   # type: ignore
+           return g.filemanager    # type: ignore
+
+       @wraps(FileManagementService.get_upload_status)
+       def get_upload_status(upload_id: int) -> Upload:
+           '''See :meth:`FileManagementService.get_upload_status`.'''
+           return current_session().get_upload_status(upload_id)
+
```
bdc34: It seems like this could be something relatively simple:
```
def setup_web_app():
...
app['FileManagementService'] = FileManagementServiceImplXYZ( config.get(xxx)...)
```
Then to use it
```
import current_app from Flask
...
current_app.FileManagementService.get_upload(id)
```
That would give an error like ```object has no attribute 'FileManagementService'``` if the app was misconfigured.

arxiv/integration/api/status.py
erickpeirson: Yes!

erickpeirson: Python 3 keeps on giving

arxiv/integration/api/service.py
erickpeirson: Good idea; pulled this out into its own function

arxiv/integration/api/service.py
erickpeirson: oh yuck; fixed

arxiv/integration/api/service.py
erickpeirson: Great question; will reply to this and https://github.com/arXiv/arxiv-base/pull/107#discussion_r266102345 here.

Last thing first: this implementation totally allows for direct instantiation, e.g. to use the same implementation with different configs in the same app. This works as expected:

```python
fm = FileManagementServiceImplXYZ( config.get(xxx)...)
fm2 = FileManagementServiceImplXYZ( config.get(yyy)...)
fm.get_upload(id)
fm2.get_upload(id)
```

One of the goals here, that could probably use some elaboration in [this slightly stale documentation](https://arxiv.github.io/arxiv-arxitecture/crosscutting/services.html#controller-modules) is to not be importing and using Flask proxy objects (most notably ``current_app``) all over the place. In particular, a requirement for controller modules in arXiv NG apps is that they avoid use of ``current_app`` and other proxy objects, in order to cut down on side-effects and reduce the amount of mocking required to test them. That's part of why we are separating "routes" from "controllers" -- Flask integration (setting up routing, working with the ``request`` proxy, etc) gets concentrated in the routes module(s), leaving the controllers to focus on the logic of handling the request.

What that means for service integration modules is that they should be handling Flask context management themselves. This mirrors the Flask extension pattern. The goal is to be able to simply import and use the integration module in controllers, and not worry about Flask concerns like contexts.

So, specifically to the question 
> What is the advantage of this over how jinja adds a dict entry to the app?

Doing this in your controllers would not be consistent with those goals:

```python
from flask import current_app

def handle_some_request(params, etc):
    current_app.someService.do_something()

```

Indeed, that's not very helpful if we are truly using the same implementation to talk to more than one service with the same API in the same app. I'd be interested to hear if there are cases where we need to do this.

arxiv/integration/meta.py
erickpeirson: See https://github.com/arXiv/arxiv-base/pull/107#discussion_r266385406

arxiv/integration/api/service.py
bdc34: I'm for the goal of removing the flask current_app and request imports.

A simpler the way to achieve these goals:

1. app.py create_web_app() sets up app with services from config or env vars or somewhere else.

2. In routes.py the blueprint code provides controllers services. The blueprint code will get called to handle the route, it will then get the service from the app dict, then it will call the controller with it needed services as arguments.

This satisfies the goals of not importing flask proxy objects and makes mocking trivial.

It also satisfies the goal of not binding the service implementation class in the controller.

This is also a pattern of use that all developers will be familiar with and well supported by development tools.

This pattern has the benefit of no code to maintain. 

route.py
```python
@blueprint.route('abs/<path:arxiv_id>', methods=['GET'])
def abstract(arxiv_id: str) -> Response:
    """Abstract (abs) page view."""

    abs_doc_service = current_app.abs_doc_service
    response, code, headers = abs_page.get_abs_page(abs_doc_service, arxiv_id)

    if code == status.HTTP_200_OK:
        return render_template('abs/abs.html', **response), code, headers
```

The blueprint is responsible to provide the abs_page controller function access to any services it needs.  The blueprint/route seems like a good place to do it because it is already deep in the territory of flask proxy imports.

erickpeirson: Yeah, dependency injection might be a reasonable way to approach this. Since this would be a departure from how I've asked the team to implement controllers, routes, and services heretofore, we'll need to be deliberate and systematic about making such a change.

If you'd like to put together a brief (one-page) summary outlining the proposed change, we can get input from the team, which will help to inform a final decision. For current on-deck releases (submission-0.1, browse-0.2, probably browse-0.3) we'll proceed as I've outlined above.

bdc34: I'm proposing something more along the lines of  "threading the state" than capital D capital I Dependency Injection.

erickpeirson: Sure, fair enough

PR #108, opened 2019-03-18
ARXIVNG-1375 support for e-mail, including templates

erickpeirson:
This wires up the base e-mail templates so that they can be used in services, adds a plaintext version of the abs macro, and some template filters to support it.


erickpeirson:
Raised this against ARXIVNG-1970 since it was branched from there

PR #109, opened 2019-03-21
Fixes bug "Links for Multiple DOIs on abs page are garbled"

bdc34:
First, the abs macro templates were expecting a single DOI were explicitly using the a DOI filter and a click through filter. 

Second, when the abs macro template was switched to the links.py urlize, the DOI text was "this  URL"

That was due to the bleach linkifier applying all of the attribute callbacks. The external ref callback is changing the _text to "this URL"

So this PR refactors link.py to add a kind -> callbacks Dict.

link.py is also refactored to no longer produce one bleach linkifier object per call to link.urlizer() but to return a function composed of each kind of linkifier. Each kind of linkifier is configured independently by as patterns and callbacks.

This PR also adds nofollow and noopener to external links.

### DIFF
--- a/arxiv/base/urls/tests/test_urlize.py
+++ b/arxiv/base/urls/tests/test_urlize.py
@@ -239,5 +239,35 @@ class TestURLize(unittest.TestCase):
     def test_arxiv_prefix(self):
         self.assertEqual(
             links.urlize("see arxiv:1201.12345"),
-            'see <a class="link-https" data-arxiv-id="1201.12345" href="https://arxiv.org/abs/1201.12345">arXiv:1201.12345</a>'
-        )
+            'see <a class="link-https" data-arxiv-id="1201.12345" href="https://arxiv.org/abs/1201.12345">arXiv:1201.12345</a>')
+
+
+    @mock.patch(f'{links.__name__}.clickthrough')
+    def test_doi_2(self, mock_clickthrough):
+        mock_clickthrough.clickthrough_url = lambda x: x
+        self.assertRegex(links.urlize('10.1088/1475-7516/2018/07/009'),
+                         r'<a.*href="https://.*10.1088/1475-7516/2018/07/009".*>10.1088/1475-7516/2018/07/009</a>')
+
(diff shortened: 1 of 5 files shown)


arxiv/base/urls/tests/test_urlize.py
```diff
@@ -197,47 +197,77 @@ def test_parens(self):
...
-            'see <a class="link-https" data-arxiv-id="1201.12345" href="https://arxiv.org/abs/1201.12345">arXiv:1201.12345</a>'
-        )
+            'see <a class="link-https" data-arxiv-id="1201.12345" href="https://arxiv.org/abs/1201.12345">arXiv:1201.12345</a>')
+
+
+    @mock.patch(f'{links.__name__}.clickthrough')
+    def test_doi_2(self, mock_clickthrough):
+        mock_clickthrough.clickthrough_url = lambda x: x
+        self.assertRegex(links.urlize('10.1088/1475-7516/2018/07/009'),
+                         r'<a.*href="https://.*10.1088/1475-7516/2018/07/009".*>10.1088/1475-7516/2018/07/009</a>')
```
mhl10: I'm curious about the wildcards here:

`r'<a.*href="https://.*10.1088/1475-7516/2018/07/009".*>10.1088/1475-7516/2018/07/009</a>')`

First wildcard is intended to match against the `class` (and any other) attribute?
Last wildcard is intended to match against `rel` (and any other) attribute?
Is the one after the protocol meant to match against clickthrough linking?

bdc34: In this test I don't really care about the class or rel. This is just testing the href and the text.
So I'm not including those in what gets tested by throwing in the wildcards.

The `mock_clickthrough.clickthrough_url = lambda x: x` part sets the clickthrough function to the identity function because I'm not interested in testing the CT in this test either.

The wildcard after the protocol was meant to skip dx.doi.org.

mhl10: 👍

PR #110, opened 2019-04-01
Pre-release merge for v0.15.3

mhl10:
The purpose of this release is to provide a fix for URL/DOI linking from ARXIVNG-1951; it will be incorporated with forthcoming browse release 0.2.1.

I'm not sure why this PR is (still) showing so many commits; it will make more sense to focus on the files themselves.


arxiv/base/templates/base/macros.html
```diff
@@ -170,7 +170,7 @@ <h1 class="title mathjax"><span class="descriptor">Title:</span>{{ title|tex2utf
         {%- if doi %}
         <tr>
           <td class="tablecell label"><abbr title="Digital Object Identifier">DOI</abbr>:</td>
-          <td class="tablecell msc_classes"><a href="{{ doi|url_for_doi|clickthrough_url }}">{{ doi }}</a></td>
+          <td class="tablecell msc_classes"><a href="{{ doi| urlize }}">{{ doi }}</a></td>
```
erickpeirson: very minor: whitespace to right of pipe

```suggestion
          <td class="tablecell msc_classes"><a href="{{ doi|urlize }}">{{ doi }}</a></td>
```

mhl10: thanks, good catch!

PR #112, opened 2019-04-09
add new Matomo tracking code

eawoods:
Whenever it's time to pull this in -- here is the tracking code for the new Matomo.
I checked with Chris and he said that it is OK to have the new code implemented on some pages and the old code on others if we want to or need to; the log import would cover that. He did say, however, NOT to have both trackers running on the same page so that it doesn't get double counted.

For the donate button tracking, this needs to get implemented for arxiv-docs specifically.
The new Matomo has other options if we need to configure download formats to track or any other custom variables. I tried to just match what we had in the old Piwik for now.

### DIFF
--- a/arxiv/base/templates/base/base.html
+++ b/arxiv/base/templates/base/base.html
@@ -8,7 +8,7 @@
   </head>
   <body>
   {% import "base/macros.html" as macros %}
-  <noscript><img src="//webanalytics.library.cornell.edu/piwik.php?idsite=538&rec=1" style="border:0;" alt="" /></noscript>
+  <noscript><p><img src="https://webstats.arxiv.org/matomo.php?idsite=1&amp;rec=1" style="border:0;" alt="" /></p></noscript>
   <header>
     {% block header %}
     {% include "base/header.html" %}
--- a/arxiv/base/templates/base/head.html
+++ b/arxiv/base/templates/base/head.html
@@ -43,14 +43,20 @@
 <script src='//static.arxiv.org/MathJax-2.7.3/MathJax.js'></script>
 
 <script src="{{ url_for('base.static', filename='js/notification.js') }}"></script>
-<!-- Piwik -->
+<!-- Matomo -->
 <script type="text/javascript">
-var _paq = _paq || [];
-_paq.push(["setDomains", ["*.arxiv.org"]]);
-_paq.push(['trackPageView']);
-_paq.push(['enableLinkTracking']);
-(function()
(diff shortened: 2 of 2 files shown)


arxiv/base/templates/base/base.html
```diff
@@ -8,7 +8,7 @@
   </head>
   <body>
   {% import "base/macros.html" as macros %}
-  <noscript><img src="//webanalytics.library.cornell.edu/piwik.php?idsite=538&rec=1" style="border:0;" alt="" /></noscript>
+  <noscript><p><img src="https://webstats.arxiv.org/matomo.php?idsite=1&amp;rec=1" style="border:0;" alt="" /></p></noscript>
```
mhl10: I'm mildly curious (but not concerned) about the `p` tag here :-)

eawoods: Yeah, I looked at that too - this was stock code from Matomo.
My best guess is that it is for strict HTML validators, after a little bit of research.
Won't hurt, it doesn't generate unless JavaScript is disabled - might be worth checking to see if an auto-margin gets applied on the p tag with a zero-size image but I suspect that it won't.

erickpeirson: Any chance we can make this (and maybe the domains/URLs further down) configurable? Would save having to mint new software versions to change this in the future. Not that it should change often, but still.

eawoods: I'd argue that for this particular case alone, setting the domain for Matomo in a config variable may not be the best choice - it is highly unlikely to change and may have as many subtle cons as pros. Would obfuscate the URL for future developers and make the tracker harder to find, increasing possibility for something to go wrong in config and take down tracking without meaning to, ending up with a data loss.

It's a minor point, and if there's a strong reason to use config that would be fine too.

bdc34:
Will matomo be disabled somehow on systems like beta and dev?

eawoods:
That's a good point @bdc34 - it should be, and I forgot that we need to include a means for turning it off for pages like login too.

bdc34:
> That's a good point @bdc34 - it should be, and I forgot that we need to include a means for turning it off for pages like login too.

Since it is an image it isn't a security concern on the login page. So if we want to track visits to the login page I think it is fine. 

Loading JS that we do not control is the security concern, especially on the login page. The image shouldn't be able to do anything, JS can do just about everything.

eawoods:
Well, that's the thing - we have both. The image is a no-JS fallback and we also have JS code in the header. Are you saying that a better method would be to just use the image and no JS at all? Are there any downsides to that? Sounds like using just the image might be the way to go, and restrict from beta/dev in a different way.

bdc34:
Liz came by and we talked about this. I think image matomo tracking but no JS on /login would be great. JS matomo tracking for the rest of the site sounds fine. We talked about a couple ways this could be achieved but didn't come to any conclusions about which to use.

PR #113, opened 2019-04-09
DOI linking regression 

bdc34:
Fixes regression with DOIs in base by adding a specialized urlizer kind for the abs DOI field.

This will be needed for browse 0.2.1

https://arxiv-org.atlassian.net/browse/ARXIVNG-2049

### DIFF
--- a/arxiv/base/urls/links.py
+++ b/arxiv/base/urls/links.py
@@ -67,8 +67,16 @@ Pattern for matching DOIs.
 We should probably match DOIs first because they are the source of a
 lot of false positives for arxiv matches.
 
-Only using the most general express from
+Only using the most general expression from
 https://www.crossref.org/blog/dois-and-matching-regular-expressions/
+"""
+
+
+BROAD_DOI=re.compile(r'(?P<doi>10\.\d{4,5}\/\S+)', re.I)
+"""
+Very broad pattern for matching DOIs in the abs DOI field.
+
+Ex. 10.1175/1520-0469(1996)053<0946:ASTFHH>2.0.CO;2
 """
 
 ARXIV_PATTERNS = [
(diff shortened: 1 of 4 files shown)


arxiv/base/urls/links.py
```diff
@@ -292,9 +308,12 @@ def urlize(text: str, kinds: List[str] = SUPPORTED_KINDS) -> str:
...
 
 
-def urlizer(kinds: List[str] = SUPPORTED_KINDS) -> Callable_Linker:
-    """
-    Generate a function to convert tokens to links.
+def urlizer(kinds: List[str] = DEFAULT_KINDS) -> Callable_Linker:
+    """Generate a function to convert tokens to links.
+
+    If the urlizing funciton is going to be reused, this is more
+    efficent than urlize because this will only call _get_linker()
```
mhl10:
```suggestion
    efficient than urlize because this will only call _get_linker()
```

arxiv/base/urls/links.py
```diff
@@ -292,9 +308,12 @@ def urlize(text: str, kinds: List[str] = SUPPORTED_KINDS) -> str:
     return _get_linker(kinds)(text)
 
 
-def urlizer(kinds: List[str] = SUPPORTED_KINDS) -> Callable_Linker:
-    """
-    Generate a function to convert tokens to links.
+def urlizer(kinds: List[str] = DEFAULT_KINDS) -> Callable_Linker:
+    """Generate a function to convert tokens to links.
+
+    If the urlizing funciton is going to be reused, this is more
```
mhl10:
```suggestion
    If the urlizing function is going to be reused, this is more
```

arxiv/base/urls/tests/test_urlize.py
```diff
@@ -271,3 +271,45 @@ def test_double_doi(self, mock_clickthrough):
                          'Should handle two DOIs in a row correctly')
 
 
+    @mock.patch(f'{links.__name__}.clickthrough')
+    def test_broad_doi(self, mock_clickthrough):
```
mhl10: This looks good!

PR #116, opened 2019-04-17
ARXIVNG-2073 add support for relative static paths

erickpeirson:
In some cases, we want an app to handle all of its static files, e.g. when deploying in a development environment. Setting the configuration param ``RELATIVE_STATIC_PATHS`` to True will cause ``/{RELATIVE_STATIC_PREFIX}`` to be prepended to the static paths for base assets. This should have no impact on static paths for blueprints.

This helps to solve the problem that static URLs don't work when a site is deployed as a container on a sub path.

### DIFF
--- a/Pipfile
+++ b/Pipfile
@@ -10,10 +10,11 @@ pytz = "*"
 uwsgi = "*"
 "boto3" = "==1.9.108"
 wtforms = "*"
-"e1839a8" = {path = "."}
+e1839a8 = {path = "."}
 bleach = "*"
 "flask-s3" = "*"
 "backports-datetime-fromisoformat" = "==1.0.0"
+arxiv-base = {path = "."}
 
 [dev-packages]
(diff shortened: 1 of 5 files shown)


Pipfile
```diff
@@ -10,10 +10,11 @@ pytz = "*"
 uwsgi = "*"
 "boto3" = "==1.9.108"
 wtforms = "*"
-"e1839a8" = {path = "."}
+e1839a8 = {path = "."}
 bleach = "*"
 "flask-s3" = "*"
 "backports-datetime-fromisoformat" = "==1.0.0"
+arxiv-base = {path = "."}
```
mhl10: Was this intentional?

erickpeirson: whoops, nope

erickpeirson:
> Did you intend to raise this against master?

Whoops, nope. Changed to develop

PR #122, opened 2019-04-23
ARXIVNG-2042 Make retry behavior in base Kinesis consumer easier to modify

erickpeirson:
The Kinesis consumer has some retry handling, e.g. when waiting for the stream to be available. https://github.com/arXiv/arxiv-base/tree/feature/ARXIVNG-1970/arxiv/integration/kinesis 

This is currently not configurable; it should be easy to alter this in a subclass without overriding whole methods. 

This PR eliminates the simple ``retry`` decorator in that module, and uses [``retry``](https://pypi.org/project/retry/) instead. Parameters for retry can be passed to the ``BaseConsumer`` constructor, and are stored in the public attribute ``retry_params``. So the retry behavior can be configured via the app config (``KINESIS_RETRY_x`` params), or a subclass can easily intervene on  ``retry_params``. 

Not perfect. But an improvement.

### DIFF
--- a/Pipfile
+++ b/Pipfile
@@ -13,6 +13,7 @@ wtforms = "*"
 bleach = "*"
 "flask-s3" = "*"
 "backports-datetime-fromisoformat" = "==1.0.0"
+retry = "*"
 
 [dev-packages]
 pydocstyle = "*"
--- a/arxiv/integration/kinesis/consumer/__init__.py
+++ b/arxiv/integration/kinesis/consumer/__init__.py
@@ -189,7 +189,7 @@ class BaseConsumer(object):
             'delay': delay,
             'max_delay': max_delay,
             'backoff': backoff,
-            'jitter': jitter
+            'jitter': jitter  #  extra seconds added to delay between retry attempts.
         }
 
         if not self.stream_name or not self.shard_id:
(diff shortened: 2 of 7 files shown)


Pipfile
```diff
@@ -13,6 +13,7 @@ wtforms = "*"
 bleach = "*"
 "flask-s3" = "*"
 "backports-datetime-fromisoformat" = "==1.0.0"
+retry = "*"
```
mhl10: ah, of course there's a package for that :-)

arxiv/integration/kinesis/consumer/__init__.py
```diff
@@ -207,64 +173,76 @@ def __init__(self, stream_name: str = '', shard_id: str = '',
...
-                'No stream indicated; making no attempt to connect to Kinesis'
-            )
-            return
+        # Retry parameters.
+        self.retry_params = {
+            'tries': tries,
+            'delay': delay,
+            'max_delay': max_delay,
+            'backoff': backoff,
+            'jitter': jitter
```
mhl10:
```suggestion
            'jitter': jitter  #  extra seconds added to delay between retry attempts.
```

erickpeirson:
> In a few cases, the `@retry` decorator might work equally well and be more clear to the reader, but `retry_call` is also fine.

Yeah, I went back and forth on this. Ended up going with ``retry_call`` so that I could handle the exceptions (if retry fails) within ``wait_for_stream``, rather than having another decorated method.

PR #125, opened 2019-04-30
Fixes race condition due to multithread use of bleach

bdc34:
This is a bit of an odd one. There were some strange errors in the production logs that were persistent but not immediately reproducible.
These turned out to be related to bleach.  https://bleach.readthedocs.io/en/latest/clean.html

https://arxiv-org.atlassian.net/browse/ARXIVNG-2107

I reproduced this on my laptop and fixed it by making the bleach object thread local. 

I can reproduce error on my laptop by:
- running apache,
- with wsgi with 16 threads,
- configiuring arxiv-browse to not use the DB
- then running "ab -c 300 -t 120 http://localhost/test/abs/1411.4413"

Erick and Jamie I'm wondering your thoughts on:
1. Are thread locals a good way to solve this problem?
2. Is the dynamic creation of the thread local bleach objects in _deferred_thread_local_linker_of_kind a good way to implement this solution? (Or is there a more straightforward way?)
3. Is this going to create any new problems?

### DIFF
--- a/arxiv/base/urls/links.py
+++ b/arxiv/base/urls/links.py
@@ -313,7 +313,12 @@ def _deferred_thread_local_linker_of_kind(kind: str) -> Callable_Linker:
         # This must not be called while the app is starting up, only
         # when the linkiers are being used in requests. So the
         # function is created/cached at call time.
-        return g.get('linkers',{}).get(kind,_get_linker_of_kind(kind))(instr)
+        if 'linkers' not in g:
+            g.linkers = {}
+        if kind not in g.linkers:
+            g.linkers[kind] = _get_linker_of_kind(kind)
+
+        return g.linkers[kind](instr)
 
     return deferred
 
--- a/arxiv/base/urls/links.py
+++ b/arxiv/base/urls/links.py
@@ -1,5 +1,5 @@
 r"""
-Patterns and functions to detect arXiv ids and Urls in text.
+Patterns and functions to detect arXiv ids and URLs in text.
 
 Functions to detect arXiv ids, URLs and DOI in text.
 Functions to transform them to <a> tags.
@@ -19,8 +19,8 @@ HTTP URLs:      http://something.org/myPaper/1234.12345
 FTP URLs:       ftp://example.com/files/1234.12345
 
 Just matching for arXiv ids with \d{4}\.\d{4,5} will match several of
-these. To deal with this we are priortizing the matches and
-interupting once one is found.
+these. To deal with this we are prioritizing the matches and
+interrupting once one is found.
 
 We should probably match DOIs first because they are the source of a
 lot of false positives for arxiv matches.
@@ -33,6 +33,7 @@ handle FTP addresses.
 Updated 20 March, 2019: refactored to provide independent bleach attribute
 callbacks for each kind of link.
 """
+from threading import local
 from typing import List, Pattern, Tuple, Callable, Dict, Union
 import re
 from functools import reduce
@@ -64,7 +65,7 @@ DOI = re.compile(   # '10.1145/0001234.1234567'
 Pattern for matching DOIs.
 
 We should probably match DOIs first because they are the source of a
-lot of false positives for arxiv matches.
+lot of false positives for arxiv id matches.
 
 Only using the most general expression from
 https://www.crossref.org/blog/dois-and-matching-regular-expressions/
@@ -90,7 +91,7 @@ ARXIV_ID = re.compile(
 )
 
 OKCHARS = r'([a-z0-9,_.\-+~:]|%[a-f0-9]*)'
-"""Chacters that are acceptable during PATH, QUERY and ANCHOR parts"""
+"""Characters that are acceptable during PATH, QUERY and ANCHOR parts"""
 
 PATH = rf'(?P<PATH>(/{OKCHARS}*)+)?'
 """Regex for path part of URLs for use in urlize"""
@@ -145,7 +146,7 @@ def _extend_class_attr(attrs: Attrs, new_class: str) -> Attrs:
 
 
 def _add_rel(attrs: Attrs, new: bool = False) -> Attrs:
-    """Check if href is interal or external and adds attrs as appropriate."""
+    """Check if href is internal or external and adds attrs as appropriate."""
     o = urlparse(attrs[(None, 'href')])
     if not o.netloc.split(':')[0].endswith('arxiv.org'):   # External link?
         return _add_rel_external(attrs, new)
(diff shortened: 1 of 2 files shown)


JaimieMurdock:
Good sleuthing! I found the [intro to ``threading` at Python Module of the Week](https://pymotw.com/2/threading/#thread-specific-data) to be helpful in understanding some of the constructs.

> 1. Are thread locals a good way to solve this problem?
I think there's two ways, the thread locals seem to do the job and keep the data partitioned with each instance of the Linker being unique to the particular thread.

> 2. Is the dynamic creation of the thread local bleach objects in _deferred_thread_local_linker_of_kind a good way to implement this solution? (Or is there a more straightforward way?)

I suspect that the non-thread-safe module - `bleach` - might be guaranteed safety by locking all your access to `bleach`'s Linker with some sort of synchronization primitive - [a `Lock` on bleach might work](https://pymotw.com/2/threading/#controlling-access-to-resources).

Although, the library does suggest for making the `Cleaner` thread-safe, using new instances per thread, as you did with locals

> 3. Is this going to create any new problems?

I'm not sure if state is shared between the instances, especially since you're got `callbacks=callbacks[kind]` as another reference - might be a possible source of a thread leak?

I've been trying to wrap my head around the Flask thread model too, and I think that's where the issues are arising. Rather than using an all-out `locals()` call, simply creating a new thread might be enough to get the safety you need. Flask itself does use `locals()` in implementation, as noted [in the design doc](http://flask.pocoo.org/docs/0.12/design/#thread-locals).

bdc34:
Jaimie,
Thanks for taking time to check this out.

> I suspect that the non-thread-safe module - `bleach` - might be guaranteed safety by locking all your access to `bleach`'s Linker with some sort of synchronization primitive - [a `Lock` on bleach might work](https://pymotw.com/2/threading/#controlling-access-to-resources).

I was hoping to not use a lock since I don't actually want to share data. 
 
> I'm not sure if state is shared between the instances, especially since you're got `callbacks=callbacks[kind]` as another reference - might be a possible source of a thread leak?

I'm hoping that will be okay since is is only called in  _get_linker_of_kind() which is put off by _deferred_thead_local_linker_of_kind() until jinja templates are calling these linkers from threads. These objects will be cached in the dict on the thread.local _t_local_linkers. 

> I've been trying to wrap my head around the Flask thread model too, and I think that's where the issues are arising. Rather than using an all-out `locals()` call, simply creating a new thread might be enough to get the safety you need. Flask itself does use `locals()` in implementation, as noted [in the design doc](http://flask.pocoo.org/docs/0.12/design/#thread-locals).

I'm glad to see they are using thread.local() because I think links.py is now trying to achieve the same sort of affect.

arxiv/base/urls/links.py
```diff
@@ -33,6 +33,7 @@
 Updated 20 March, 2019: refactored to provide independent bleach attribute
 callbacks for each kind of link.
 """
+from threading import local
```
erickpeirson:
```suggestion
from flask import g
```

arxiv/base/urls/links.py
```diff
@@ -302,11 +302,44 @@ def _compose_list_of_funcs(fv: List[Callable_Linker]) -> Callable_Linker:
...
+    #
+    # This will return a Callable that creates the bleach linkifier at
+    # only when called with the string to linkify. This is so the bleach
+    # linkifier can be put into a thead.local of each individual WSGI thread.
+
+    def deferred(instr: str) -> str:
+        # This must not be called while the app is starting up, only when the
+        # linkiers are being used in threads. So the function is created/cached
+        # at call time.
+        global _t_local_linkers
```
erickpeirson: L321-328 could be something like:

```python
if linkers not in g:
    g.linkers = {}
if kind not in g.linkers:
    g.linkers[kind] = _get_linker_of_kind(kind)

return g.linkers[kind]
```

erickpeirson:
> > I've been trying to wrap my head around the Flask thread model too, and I think that's where the issues are arising. Rather than using an all-out `locals()` call, simply creating a new thread might be enough to get the safety you need. Flask itself does use `locals()` in implementation, as noted [in the design doc](http://flask.pocoo.org/docs/0.12/design/#thread-locals).
> 
> I'm glad to see they are using thread.local() because I think links.py is now trying to achieve the same sort of affect.

Just to speak directly to this point. From those docs: "Flask uses thread local objects (**context local objects in fact, they support greenlet contexts as well**)" -- it's a bit misleading, but they are not just using ``thread.local()``. That's a reference to ``g``'s use of Werkzeug's ``LocalStack``.

arxiv/base/urls/links.py
bdc34: It seems like urlize is just a utility to convert strings to other strings. It is kind of distressing that it then would be dependent on Flask and Werkzeurg. Maybe the setup to do the late creation of the bleach objects should be up in the Flask factory instead of in links.py?

erickpeirson: > It is kind of distressing that it then would be dependent on Flask and Werkzeurg.

The ``arxiv.base.urls`` module is already [dependent on Werkzeug and Flask](https://github.com/arXiv/arxiv-base/blob/master/arxiv/base/urls/__init__.py#L64), so this doesn't introduce a new dependency.

> Maybe the setup to do the late creation of the bleach objects should be up in the Flask factory instead of in links.py?

Not sure that I follow. Do you mean initializing ``g.linkers`` in [``arxiv.base.Base``](https://github.com/arXiv/arxiv-base/blob/master/arxiv/base/__init__.py#L38)?

The point of Flask's proxy objects is to provide thread- and context-safe extensibility for applications without having to wire everything up tightly together. This use-case is precisely what ``g`` is designed for.

arxiv/base/urls/links.py
```diff
@@ -302,11 +301,33 @@ def _compose_list_of_funcs(fv: List[Callable_Linker]) -> Callable_Linker:
...
+    # This will return a Callable that creates the bleach linkifier
+    # only when called with the string to linkify. This is so the
+    # bleach linkifier can be put into a Flask.g of each
+    # individual WSGI thread.
+
+    def deferred(instr: str) -> str:
+        # This must not be called while the app is starting up, only
+        # when the linkiers are being used in requests. So the
+        # function is created/cached at call time.
+        return g.get('linkers',{}).get(kind,_get_linker_of_kind(kind))(instr)
```
erickpeirson: This looks good. Not seeing where you set ``g.linkers`` though? Without that it looks like this will just create a new ``bleach.linkifier.Linker`` each time.

bdc34: Oh ya.

PR #127, opened 2019-05-10
"arxiv.org" should not be generically urlized with "this http url"

bdc34:
See abstract in https://arxiv.org/abs/1806.04641 

It can be urlized, but should preserve the text "arXiv.org" with link to https://arxiv.org 

Linking to arXiv IDs to should not be affected.

### DIFF
--- a/arxiv/base/urls/tests/test_urlize.py
+++ b/arxiv/base/urls/tests/test_urlize.py
@@ -360,8 +360,14 @@ class TestURLize(unittest.TestCase):
         with self.app.app_context():
             urlize = links.urlizer()
 
-            self.assertNotRegex(urlize('somethign https://arxiv.org bla'), r'this.*URL',
+            self.assertNotRegex(urlize('something https://arxiv.org bla'), r'this.*URL',
                              'arxiv.org should not get "this https URL" ARXIVNG-2130')
 
-            self.assertNotRegex(urlize('somethign arxiv.org bla'), r'this.*URL',
+            self.assertNotRegex(urlize('something arxiv.org bla'), r'this.*URL',
                                 'arxiv.org should not get "this https URL" ARXIVNG-2130')
+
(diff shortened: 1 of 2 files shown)


arxiv/base/urls/tests/test_urlize.py
```diff
@@ -347,3 +355,13 @@ def test_dont_urlize_category_name(self):
                              'category name math.CO should not get urlized')
             self.assertIn('href="http://supermath.co', urlize('supermath.co'),
                           'hostname close to category name should get urlized')
+
+    def test_dont_urlize_arxiv_dot_org(self):
+        with self.app.app_context():
+            urlize = links.urlizer()
+
+            self.assertNotRegex(urlize('somethign https://arxiv.org bla'), r'this.*URL',
```
mhl10:
```suggestion
            self.assertNotRegex(urlize('something https://arxiv.org bla'), r'this.*URL',
```

arxiv/base/urls/tests/test_urlize.py
```diff
@@ -347,3 +355,13 @@ def test_dont_urlize_category_name(self):
...
+                             'arxiv.org should not get "this https URL" ARXIVNG-2130')
+
+            self.assertNotRegex(urlize('somethign arxiv.org bla'), r'this.*URL',
```
mhl10:
```suggestion
            self.assertNotRegex(urlize('something arxiv.org bla'), r'this.*URL',
```

mhl10: Might be nice to have a test that ensures the original text and casing is preserved, e.g.
```python
urlized_arxiv = urlize('on the open-access server arXiv.org we find...')
self.assertNotRegex(urlized_arxiv, r'this.*URL', 'arXiv.org should not get "this https URL" [ARXIVNG-2130]')
self.assertRegex(urlized_arxiv, r'arXiv\.org', 'arXiv.org should be preserved as text [ARXIVNG-2130]')

```

PR #132, opened 2019-05-31
Disable auto-linking or urlization in abstract

ibnesayeed:
In response to https://twitter.com/ibnesayeed/status/1133908420099100672 thread a ticket was created at https://arxiv-org.atlassian.net/browse/ARXIVNG-2180. This PR attempts to fix this issue by disabling auto-linking in abstract. However, if auto linking is desired, then the filter needs to be improved to preserve the original text as the anchor text. I believe that people usually do not add very long URLs in abstract (unlike other places such as comments), so masking the original text in abstract is more problematic than helpful. Another side-effect of auto-linking is occasional false-positives such as realizing ASP.Net as a URL.

Replacing original text with a generic anchor text may yield something as shown below.

arxiv-abstract-this-http-url

### DIFF
--- a/arxiv/base/templates/base/macros.html
+++ b/arxiv/base/templates/base/macros.html
@@ -138,7 +138,7 @@
     <h1 class="title mathjax"><span class="descriptor">Title:</span>{{ title|tex2utf|arxiv_id_urlize|safe }}</h1>
     <div class="authors"><span class="descriptor">Authors:</span>{{ authors }}</div>
     <div class="dateline">{{ abs_date_line(arxiv_id, submitted_date, version, submission_history) }}</div>
-    <blockquote class="abstract mathjax"><span class="descriptor">Abstract:</span>{{ abstract|tex2utf_no_symbols|abstract_lf_to_br|urlize|safe }}</blockquote>
+    <blockquote class="abstract mathjax"><span class="descriptor">Abstract:</span>{{ abstract|tex2utf_no_symbols|abstract_lf_to_br|safe }}</blockquote>
     <!--CONTEXT-->
     <div class="metatable">
       <table summary="Additional metadata">


erickpeirson:
@ibnesayeed Awesome, excited to see this PR! We tossed this around quite a bit today, and decided on the following approach (rather than just disabling entirely):
- [ARXIVNG-2180](https://arxiv-org.atlassian.net/browse/ARXIVNG-2180) - Retain original link text when converting URLs to links in abstract
- [ARXIVNG-2182](https://arxiv-org.atlassian.net/browse/ARXIVNG-2182) - Do not convert URL-like tokens in abstracts to links if they do not have an explicit scheme
- [ARXIVNG-2183](https://arxiv-org.atlassian.net/browse/ARXIVNG-2183) - When a token is converted to a link in an abstract, the displayed text should be shortened if greater than 30 characters in length.

We've tentatively added this to the next minor release (base v0.16.1/browse v0.3.1), so can get to production fairly quickly.

I believe that all of these can be implemented by adding appropriate callbacks here: https://github.com/arXiv/arxiv-base/blob/master/arxiv/base/urls/links.py#L274. 

If this looks like something you want to tackle in part or whole, we're grateful for the help, and can leave this PR open for that purpose. In any case, appreciate your input on this issue.

ibnesayeed:
Thanks @erickpeirson for the feedback. I also saw the conversation on Atlassian. I support ARXIVNG-2180 and ARXIVNG-2182, but I would vote against ARXIVNG-2183. The reason is, I still believe people often do not place very long URLs in an abstract (do we have any example otherwise?), because more than the HTML, it would hurt the PDF rendering in the first place. And if someone does include a long URL, it perhaps will not be an opaque string and might contain useful information both in the domain name and towards the end of the path or query parameters. Truncating it at a hard boundary might yield in some unfortunate or less useful phrases. Wrapping issue can be handled using CSS in more than one ways (if only we have a real example to test) and if we must truncate, we can use CSS `text-overflow: ellipsis` so the original text can still be copied when highlighted while truncating it visually.

I can take care of some of these, but unfortunately, I will be travelling the whole next week for a conference, hence I cannot promise.

erickpeirson:
Sounds good. Grateful for anything that you can do.

We'll toss around the truncate vs. overflow question a bit more. As you say, it will be helpful to work with some examples.

bdc34:
Thanks but we won't be using this.

Issue #138, opened 2019-07-11
Improve Docker image for efficiency and size


ibnesayeed:
I suggest the following changes to improve the Dockerfile of this repo:

* [ ] Remove `/tmp/safe_yum.sh` file after it is executed
* [ ] Move environment variable declarations and other commands that do not change often above frequently changing layers for better caching
* [ ] Combine multiple `ENV` declarations in one to minimize layers

erickpeirson:
I'll add to this:

- [ ] Since arxiv-base package requires mysql, we should add ``mysql mysql-devel`` to the yum install step.
- [ ] For images that extend ``arxiv/base`` (e.g. for a particular service), we should start putting pip-related commands on a single step, ending with ``&& rm -rf ~/.cache/pip``. For example:

```docker
RUN pip install -U pip pipenv uwsgi \
    && pipenv install \
    && rm -rf ~/.cache/pip
```

ibnesayeed:
> Since arxiv-base package requires mysql, we should add mysql mysql-devel to the yum install step.

Shouldn't we keep MySQL in a separate container (as a microservice), utilizing the official image?

erickpeirson:
> > Since arxiv-base package requires mysql, we should add mysql mysql-devel to the yum install step.
> 
> Shouldn't we keep MySQL in a separate container (as a microservice), utilizing the official image?

IIRC, the problem is that the ``mysqlclient`` Python package expects at least the MySQL client to be installed locally

ibnesayeed:
> IIRC, the problem is that the `mysqlclient` Python package expects at least the MySQL client to be installed locally

It looks like [the package](https://pypi.org/project/mysqlclient/) requires Python and MySQL development headers and libraries, but not the client itself (unless I missed something).

erickpeirson:
> It looks like [the package](https://pypi.org/project/mysqlclient/) requires Python and MySQL development headers and libraries, but not the client itself (unless I missed something).

Yeah, you're right. So since we're on Centos7, should be just ``mariadb-devel`` I think?

ibnesayeed:
... and `python3-devel` too I guess?

erickpeirson:
Want to take a look at https://github.com/arXiv/arxiv-base/pull/139 and see what you think so far?

ibnesayeed:
Looks like it is taking a good shape. I have posted a couple minor comments/suggestions there.

ibnesayeed:
[...]

Issue #130, opened 2019-05-29
Consider not running ``yum update`` in Dockerfile


erickpeirson:
https://github.com/arXiv/arxiv-base/blob/cc9eb8f9b5e22e643c3a669420d6f9931acb1bfd/Dockerfile#L11

See https://developers.redhat.com/blog/2016/03/09/more-about-docker-images-size/

bbarker:
Some consolation is that we are doing `yum clean` in the same RUN command

https://github.com/arXiv/arxiv-base/blob/cc9eb8f9b5e22e643c3a669420d6f9931acb1bfd/Dockerfile#L35

erickpeirson:
True. Current thoughts:
- Do we really need ``update`` or ``update --security`` when the latest build of ``centos:7`` should be up to date? e.g. see https://github.com/CentOS/sig-cloud-instance-images/blob/9a389e11460bd809bb83c5fd4f095dbc8398d8cc/docker/Dockerfile
- I tried building the accounts app without them, and saved about 100 MB. So that was pretty nice.

bbarker:
Good points. CC @rcplane - any thoughts? (I know you are sick today - so just flagging you for your thoughts at some point).

PR #139, opened 2019-07-15
ARXIVNG-2462 cleaning up Dockerfiles for performance, size

erickpeirson:
This is a start on #138 ; I also poked at #130. Looks like a modest improvement so far:

- arxiv/base: 206MB -> 177MB
- arxiv/fourohfour: 359MB -> 276MB

Incidentally, this also delivers ARXIVNG-2207 (Configure default backend to use APPLICATION_ROOT correctly).


Dockerfile
```diff
@@ -2,16 +2,17 @@
 
 FROM centos:centos7
 
-#
+WORKDIR /opt/arxiv
```
ibnesayeed: Do we need this?

erickpeirson: Maybe not. By convention we use this is most of our downstream images, so thought it would be less redundant to put it here?

fourohfour/Dockerfile
```diff
@@ -1,21 +1,13 @@
 # arxiv/fourohfour
 
-FROM arxiv/base:0.14.2
+FROM arxiv/base:ARXIVNG-2462
```
ibnesayeed: How about parametrizing it like this?

```suggestion
ARG     ARXIVNG_VER=ARXIVNG-2462
FROM arxiv/base:${ARXIVNG_VER}
```

The default value can be something more generic such as `latest`, but in the build pipeline proper `--build-arg` can be supplied to externally supply the value without the need of updating this Docker file each time a new base image is ready.

erickpeirson: Nice, I like that

Dockerfile
ibnesayeed: While this is great to move shared ideas up in the chain, unless we can guarantee that in the future no one would add another `WRKDIR` towards the end of this image, we should avoid such things. If we do choose to keep it here, should we add a comment note above/below it saying, this image should have the final `WORKDIR` set so in order to avoid derived images to break silently.

fourohfour/Dockerfile
erickpeirson: Added this in dce0639

Dockerfile
erickpeirson: Gotcha. I think it makes sense to remove it, given that there are no obvious strong drivers to only and always use ``/opt/arxiv``.

erickpeirson: Fixed in 3701e20

Dockerfile
```diff
@@ -31,16 +33,27 @@ for pkg in ${PKGS_TO_INSTALL}; do\n\
...
+
+LABEL version="0.15.9" \
+  maintainer="arXiv <nextgen@arxiv.org>" \
+  org.label-schema.schema-version="1.0" \
+  org.label-schema.name="arXiv Base" \
+  org.label-schema.description="Base image for arXiv NG applications" \
+  org.label-schema.url="https://arxiv.github.io/arxiv-base" \
+  org.label-schema.vcs-url="https://github.com/arxiv/arxiv-base" \
+  org.label-schema.vendor="arXiv.org" \
+  org.label-schema.version="0.15.9"
```
ibnesayeed: You may want to move the labels and envs that will remain unchanged up above yum install, so the layers should be preserved. Labels that change can be added towards the end. Also, if the version number is to be used in more than one places, we can declarer that as an independent ENV var and reuse that variable in different places.

fourohfour/Dockerfile
```diff
@@ -1,21 +1,27 @@
...
+
+LABEL version="0.15.9" \
+  maintainer="arXiv <nextgen@arxiv.org>" \
+  org.label-schema.schema-version="1.0" \
+  org.label-schema.name="arXiv FourOhFour" \
+  org.label-schema.description="Default error backend for arXiv k8s clusters" \
+  org.label-schema.url="https://arxiv.github.io/arxiv-base" \
+  org.label-schema.vcs-url="https://github.com/arxiv/arxiv-base" \
+  org.label-schema.vendor="arXiv.org" \
+  org.label-schema.version="0.15.9"
```
ibnesayeed: Labels that do not change from the base image should not be redeclared here.

Dockerfile
erickpeirson: Cool; moved the labels in 87eff32 -- going to just go with one version tag for now

fourohfour/Dockerfile
erickpeirson: Fixed in 87eff32

Dockerfile
```diff
@@ -31,16 +38,19 @@ for pkg in ${PKGS_TO_INSTALL}; do\n\
...
+  && rm -rf ~/.cache/pip
 
-#
-# This is needed by click: http://click.pocoo.org/5/python3/
-#
-ENV LC_ALL=en_US.UTF-8
-ENV LANG=en_US.UTF-8
+ENV LC_ALL=en_US.UTF-8 \
+    LANG=en_US.UTF-8 \
+    APPLICATION_ROOT="/"
```
ibnesayeed: We can move this `ENV` declaration up as well unless an intermediary step might override any of these values.

Dockerfile
```diff
@@ -2,16 +2,22 @@
...
 FROM centos:centos7
 
-#
+LABEL maintainer="arXiv <nextgen@arxiv.org>" \
+  org.label-schema.schema-version="1.0" \
+  org.label-schema.name="arXiv Base" \
+  org.label-schema.description="Base image for arXiv NG applications" \
+  org.label-schema.url="https://arxiv.github.io/arxiv-base" \
+  org.label-schema.vcs-url="https://github.com/arxiv/arxiv-base" \
+  org.label-schema.vendor="arXiv.org"
```
ibnesayeed: I am not too picky about this, but the [label-schema](http://label-schema.org/rc1/) is now deprecated in favor of [OCI Image Spec](https://github.com/opencontainers/image-spec/blob/master/annotations.md).

erickpeirson: Sounds good; switched in 67927e7

Dockerfile
erickpeirson: Fixed in 67927e7

ibnesayeed:
This LGTM!

Though, I do not know the purpose of `Dockerfile-with-git` file, which is still there untouched.

erickpeirson:
> Though, I do not know the purpose of `Dockerfile-with-git` file, which is still there untouched.

Oh yeah, that's no longer needed. Removed in 3d21f7a

Dockerfile
```diff
@@ -31,16 +44,15 @@ for pkg in ${PKGS_TO_INSTALL}; do\n\
...
+  && python3.6 get-pip.py \
+  && pip install -U pip pipenv uwsgi \
+  && rm -rf ~/.cache/pip
 
-#
-# This is needed by click: http://click.pocoo.org/5/python3/
-#
-ENV LC_ALL=en_US.UTF-8
-ENV LANG=en_US.UTF-8
+LABEL version="0.15.9"
```
ibnesayeed: I think even this can go above `yum` command right after ENV declarations. Because, version string is not going to change on every commit or updates in dependencies, so this layer would be wasted each time. However, it is important to keep it independent form other metadata labels.

erickpeirson: Sounds good. Fixed in f491c19

Dockerfile
```diff
@@ -21,6 +35,8 @@ PKGS_TO_INSTALL=$(cat <<-END\n\
   python36u-devel\n\
   which\n\
   wget\n\
+  mariadb-devel\n\
+  sqlite\n\
```
ibnesayeed: It is usually a good practice is to keep such lists alphabetically sorted to avoid duplicate inclusion in the future as the list grows.

erickpeirson: Good catch. Fixed in 4e47107

erickpeirson:
Working through some of our downstream images to verify that this continues to work. Will update this comment as I go.

- [arxiv/submission-ui](https://hub.docker.com/r/arxiv/submission-ui/tags): working well at https://github.com/arXiv/arxiv-submission-ui/commit/dc47358712d590b756d216abe962bd06b856dd94, reduced size from 349MB -> 254MB.

ibnesayeed:
Added some comments in the https://github.com/arXiv/arxiv-submission-ui/commit/dc47358712d590b756d216abe962bd06b856dd94 commit.

erickpeirson:
I'm happy with where this ended up, and after testing this out with several downstream service I'm comfortable that this does not break anything. We can continue to revise if other issues/opportunities arise. Going to go ahead and merge so that we can start building against this.

PR #141, opened 2019-07-15
ARXIVNG-2182 ARXIVNG-2180 ARXIVNG-2183 Preserve URLs in link label, but shorten to 30 characters or less

erickpeirson:
Currently we render URLs in abstracts as "this HTTP URL", with an anchor tag pointing at the original URL. This can be seen as contrary to submitter intent, and hides information from readers. On the other hand, it prevents potentially long and ugly URLs from being displayed. 

This PR delivers the following changes:

- ARXIVNG-2183 - When a token is converted to a link in an abstract, the displayed text should be shortened if greater than 30 characters in length
- ARXIVNG-2182 - Do not convert URL-like tokens in abstracts to links if they do not have an explicit scheme
- ARXIVNG-2180 - Retain original link text when converting URLs to links in abstract

This attempts to strike a balance between submitter intent, transparency to the end user, and not mangling the display with long URLs. When dealing with URLs greater than 30 characters in length, I have attempted to preserve as much of the hostname as possible, and include the last three characters of the path/query part.

Updated test cases to reflect these changes.


erickpeirson:
@ibnesayeed Does this look like an OK alternative to #132 ?

arxiv/base/urls/links.py
```diff
@@ -176,6 +177,53 @@ def _this_url_text(attrs: Attrs, new: bool = False) -> Attrs:
...
 
+def _remove_scheme_from_label(attrs: Attrs, new: bool = False) -> Attrs:
+    label = attrs['_text']
+    o = urlparse(label)
+    if o.scheme and f'{o.scheme}://' in label:
+        _, attrs['_text'] = label.split(f'{o.scheme}://', 1)
+    return attrs
+
+
+def _shorten_text_to_30_characters(attrs: Attrs, new: bool = False) -> Attrs:
```
ibnesayeed: I would rename it with something like `_shorten_url_to_n_characters` because it is not for any text, but a more specific type (i.e., a URL) and encoding an arbitrary number in the function name means if the limit is to be changes in future, all the function references will need a modification. This function can have an extra parameter of `max_length` which can default to `30` or whatever number we settle with.

erickpeirson: Renamed to ``_shorten_label_to_n_characters`` in c2348a9. Because of the way this is used (as a callback for bleach) we need to keep the parameterization via the config proxy. In other words, an app can override the default by setting ``MAX_URL_DISPLAY_LENGTH``.

ibnesayeed:
This looks way better than the existing state. However, personally do not think there is a need to truncate URLs in the first place. Instances of long URLs should be very rare. We should assess some real examples if they really make such a visual ugliness that we take preventive measure that would affect the rest.

Also, removing protocol from the anchor text has its own implications. For example, `web browsers show a padlock icon on https://example.com/, but not on http://example.com/` would become `web browsers show a padlock icon on example.com/, but not on example.com/`.

erickpeirson:
> If we are going to reconsider how we display URLs in the abstracts, I suggest not turning them into tags. We should drop this feature because it adds little value.

The goal is to improve the existing feature by more closely reflecting the intent of the submitter. At this time I am not aware of a reason to discontinue support for this feature.

erickpeirson:
> We should assess some real examples if they really make such a visual ugliness that we take preventive measure that would affect the rest.

I used the test ``/macros`` route in the base test app to simulate some long URLs that are representative of what we commonly see in abstracts. Here's with no truncation:

image

Looks like things wrap about as well as you'd expect. I intentionally chose a width that looks the ugliest.

Truncating to a total of 100 characters doesn't have much of an impact.

image

Truncating to 50 characters is much more tidy, but of course losing some content in the middle:

image

Truncating to 30 characters definitely makes things look the most tidy, but it's pretty hard to see what the links are really pointing to:

image

erickpeirson:
> Also, removing protocol from the anchor text has its own implications.

Fair enough. I added them back for now. It only really made sense to me to remove them if we're truncating to something fairly short (e.g. 30 characters). If we go longer or don't truncate, I see less of a reason to drop the protocol.

erickpeirson:
(going to let these tests keep failing until we arrive at a final version)

ibnesayeed:
Thanks for adding the example @erickpeirson. How about trying this CSS on the original untruncated URLs?

```
a {
  word-break: break-all;
}
```

To ensure this CSS doesn't effect other hyperlinks, limit their scope by specifying a class like `url-anchor-text`.

erickpeirson:
Nice. Added the class ``link-auto``, and sass:

```
a.link-auto
  word-break: break-all
```

image

ibnesayeed:
If the CSS solution is acceptable then perhaps we can get rid of the complex URL truncation logic. Less code is better for maintainability. However, if there is a potential use case for that elsewhere then there is no harm in keeping it around.

mhl10:
Whether or not we decide to display a truncated or un-truncated version of the original link, I think it's important we engage stakeholders on this.

If we go with truncated, I'd be favorable to 30-character truncation, but I'd just truncate the end rather than in the middle. There is some (admittedly minor) precedence for this, in how we [display](https://github.com/arXiv/arxiv-browse/blob/02ac45d21ea86e762e13d025db7b15b80cb9874f/browse/templates/tb/macros.html#L5) recent trackback links. See also [this](https://arxiv.org/).

erickpeirson:
> Whether or not we decide to display a truncated or un-truncated version of the original link, I think it's important we engage stakeholders on this.

Indeed. We're going to start with a brief consultation of end users, and go from there. Will update this thread.

erickpeirson:
Interestingly, initial reactions to a user survey indicate a strong preference to keep the existing functionality. I wonder what it would have looked like had we not included that as an option. Once we get some additional responses (currently just ~300) I'll see if there's a signal among the other three (no truncation, truncation at end, truncation in middle).

Another suggestion from a user: 

> How about instead of a truncated link show just the domain,
e.g.
>
> [www.cornell.edu]
>
> I've seen this method on other sites and it elegantly solves the problem
of showing there is a link to an external site without making it too long.

Grist for the mill.

ibnesayeed:
> Interestingly, initial reactions to a user survey indicate a strong preference to keep the existing functionality.

Responses also depend on the context and examples that were provided, if any. unless we know the intent of the author, making any changes in the original text can be surprising in some cases in one way or the other. If the author has no issues keeping the original URL in the PDF instead of using hyperlinks with custom anchor text there then why would we want to overwrite the original material. As far as auto-linking is concerned, perhaps there is no harm in doing that to aid users of the site to easily navigate to links, but without altering the visible text.

erickpeirson:
This is a really interesting question. First, a summary of the survey results:

We asked users to rank four options from most preferred to least preferred. Each option had two accompanying screenshots: a close-up of a line with a long URL, and a whole abs display with long URLs. You can see the text accompanying the four options in the results below.
image

Some take-aways:
- People seem to be fine with the way things are. That doesn't mean it's necessarily the best thing, or that they wouldn't feel that way if we had done something else all along, but here we are.
- There seems to be a pretty clear negative indication against displaying the whole URL.

Let's come back to the core drivers:
- We turn user-provided URLs into links to aid the reader in navigating to related content.
- We use a short display text in order for the abs page to not look messy. Yeah, it's down to aesthetics here (personally, I think the full URL with wrap looks just fine), but for whatever reason an aesthetic preference appears to exist. So if we do something that rubs against that aesthetic, we need to have a very clear and compelling reason.
- We recognize the need to respect submitter intent. We have discussed how the submitter is the one providing a long URL. But we may be overlooking the fact that the submitter validates a preview of the abstract display prior to finalizing their submission, and that that preview shows the substituted ``this HTTP URL`` text instead of the full URL. Obviously if we were to change this behavior going forward we would update the preview as well. But we can't overlook the fact that a change to displaying the full URL on past e-prints could be construed as contrary to the intent of the submitter.
- Using ``this HTTP URL`` as display text arguably does obscure the destination of the link, which is contrary to the interests of the reader.

With that in mind, I'm strongly leaning toward the compromise of using the domain part of the URL as the link text instead of ``this HTTP URL``. We can and should consider other ways to give submitters more control over how links are displayed. Some ideas for discussion:

- Support several different link display options (current, domain part only, full URL, don't convert to links), and let the submitter indicate their preference.
- Support limited markdown or markup in the abstract, and leave it up to the submitter to generate links.

ibnesayeed:
Thanks for sharing this @erickpeirson. Some comments below the fold.

Note: I am not sure what did the feedback questionnaire look like, so it is difficult for me to draw conclusions from answers.

> There seems to be a pretty clear negative indication against displaying the whole URL.

Did you mean people do not want the whole URL to be displayed? If that's what you meant, I don't think it is that conclusive from the plot above. For example, in the fourth group of bars it is by far the most popular choice. That said, i do not know what these numbers 1--4 represent on the X axis.

> But we may be overlooking the fact that the submitter validates a preview of the abstract display prior to finalizing their submission, and that that preview shows the substituted `this HTTP URL` text instead of the full URL. Obviously if we were to change this behavior going forward we would update the preview as well. But we can't overlook the fact that a change to displaying the full URL on past e-prints could be construed as contrary to the intent of the submitter.

I don't think we can conclude based on this argument because the submitter had no other choices to make other than perhaps deciding not to submit. I as a submitter would not withdraw my submission plan just because the displayed abstract text is not exactly what I would like while I know people can read the actual abstract from the PDF itself, though I would have preferred if my intent was not overwritten. Intent of the submitter is perhaps more clear in the PDF where the submitter had the choice to have a custom anchor text with a hyperlink but decided to go with the URL itself.

We need to think should we make a decision based on the popularity or correctness. It is generally alright to have a nice-looking hyperlink because in most cases it does not matter how the link was displayed. However, there are cases (as discussed earlier) where such decisions might render the text meaningless, confusing, misleading, or incorrect. Auto linking is done in many ways on the web, some use complete URL, some truncate it, some attempt to fetch the title of the URL if it is live and hyperlink that, and some even embed rich cards. People seem to like the last two options on the web. That does not mean these are suitable choices in ever context.

erickpeirson:
> Did you mean people do not want the whole URL to be displayed? If that's what you meant, I don't think it is that conclusive from the plot above. For example, in the fourth group of bars it is by far the most popular choice. That said, i do not know what these numbers 1--4 represent on the X axis.

1: most preferred; 4: least preferred. In other words, displaying the whole URL was selected as the least preferred option by just under 300 of the 595 respondents, and the status quo (``this HTTP URL``) was selected as the most preferred option by just over 300 respondents.

> I don't think we can conclude based on this argument because the submitter had no other choices to make other than perhaps deciding not to submit.

That's fair. Perhaps "intent" was the wrong word. I think that the salient point here is that we have shown submitters a preview, and so major changes from that preview must be considered carefully.

> We need to think should we make a decision based on the popularity or correctness.

Indeed, this is a very important point. In addition to the core drivers I mentioned above, one other worth noting here that directly informs "correctness" (for indeed, there are many ways to do this):

- We should consider the impact on accessibility. [This article about link text in WCAG](https://www.w3.org/TR/UNDERSTANDING-WCAG20/navigation-mechanisms-refs.html) is helpful. While many of the decisions that they describe are currently out of our hands (unless/until we implement markup/down support), this paragraph may be instructive:

> The text of, or associated with, the link is intended to describe the purpose of the link. In cases where the link takes one to a document or a web application, the name of the document or web application would be sufficient to describe the purpose of the link (which is to take you to the document or web application). Note that it is not required to use the name of the document or web application; other things may also describe the purpose of the link.

It seems to me that "this HTTP URL" does a less good job of describing the purpose of the link. While I'm not sure that we want to go as far as retrieving the title of the targetted page here, at least displaying the domain part may provide marginally more information about the purpose of the link.

It's important to keep in mind that this can evolve over time. It partly the nature of arXiv that our user base has an especially conservative attitude toward change, so it is important to keep changes incremental and seek feedback continuously. 

Let's start with the more limited problem of "how can we make the link text better at describing the purpose of the link, without major side-effects." I think we're starting to develop a good set of possible future features/changes, too.

ibnesayeed:
> 1: most preferred; 4: least preferred.

Thanks for clarifying this. Since I never saw the questionnaire I had trouble understanding the grouping.

> In other words, displaying the whole URL was selected as the least preferred option by just under 300 of the 595 respondents, and the status quo (`this HTTP URL`) was selected as the most preferred option by just over 300 respondents.

This is rather surprising to me and at the same time suggesting more strongly that perhaps popularity is not a good way to go in this case. Everyone involved in this conversation (here on GitHub, on Atlassian ticket, on Twitter, or people here in my research lab) have all hands down agreed that `this HTTP URL` is not how it should be. However, it turned out to be the most popular choice in the opinion poll. I am trying to understand the reason of this, could it be due to many people responding to the poll without understanding the issue (ignorance might be a harsh word), not explaining the issue well in the poll, tendency of resisting against changes, or something else?

Thanks for the WCAG link, it was an interesting read.

I understand it is important to keep all the stakeholders in the loop. One possible way is to make reasonable changes and inform them about it in the form of a blog post with some good examples in support of those changes.

To be honest, until I encountered this issue a couple months ago in one of our submissions, even I did not notice there existed such a problem, though I am using arXiv for years. I don't even recall when was this auto-linking feature introduced. This perhaps means there are not too many abstracts that have URLs in them otherwise someone would have pointed it out earlier. One way to quantify the problem would be to take a random sample of abstracts and see how many of those have 0, 1, 2, or more URLs in them. Furthermore, evaluate the distribution of lengths of those URLs (a histogram/box plot would do here). If the percentage of abstracts containing URLs is small then even a silent update (towards a more accurate representation) will not upset many people, but if this number is significant then the histogram/box plot of URL lengths will allow us to make an informed choice about truncation (or otherwise).

PR #145, opened 2019-07-29
Only disable greek textism, not letter ones

bdc34:
The fix for https://arxiv-org.atlassian.net/browse/ARXIVNG-1612 disabled tex2utf for TeX letters and greek symbols when only the greek symbols should have been disabled. 

This PR changes tex2utf() so there is a greek flag which only enables or disabled greek TeX to UTF8 translation.

The Jinja HTML filter tex2utf_no_symbols() that is used for the abstract is also changed to to disable greek but enable other letter TeXisms. 

https://arxiv-org.atlassian.net/browse/ARXIVOPS-805

### DIFF
--- a/arxiv/util/tests/test_tex2utf.py
+++ b/arxiv/util/tests/test_tex2utf.py
@@ -187,6 +187,7 @@ class TextTex2Utf(TestCase):
         self.assertEqual(utf_out, 'ARXIVDEV-2322 ' + chr(0x110) + ' fix')
 
     def test_ARXIVOPS805(self):
+        """{\aa} wasn't being converted correctly in /abs abstract field to å"""
         self.assertEqual(tex2utf("ARXIVOPS-805 \\aa  fix"),
                          'ARXIVOPS-805 å fix')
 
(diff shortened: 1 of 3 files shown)


arxiv/util/tests/test_tex2utf.py
```diff
@@ -186,18 +186,12 @@ def test_ARXIVDEV2322fixes(self):
...
-        #                   "&#x3BB;",
-        #                   'arXiv::Filters::Tex2UTF::escapeUTF8("\x{03bb}")')
-        #
-        #     latin1 = "\x{91}\x{92}\x{93}\x{94}\x{96}\x{97}\x{98}\x{A0}\x{A6}\x{B1}\x{B2}\x{B3}\x{B5}\x{BC}\x{BD}\x{BE}"
-        #     latin1expected = "`'\"\"----~ |{\\pm}^2^3{\\mu}1/41/23/4"
-        #     is (arXiv: : Filters: : Tex2UTF: : latin2tex("Test Test_String. \x{03bb} latin1" ),
-        #         "Test Test_String. {\\lambda} latin1expected",
-        #         "arXiv::Filters::Tex2UTF::latin2tex(`'\"\"----~ |{\\pm}^2^3{\\mu}1/41/23/4)")
-        #     done_testing
+    def test_ARXIVOPS805(self):
```
erickpeirson: Nice to use the ticket number here. Mind adding a brief description of what the ticket is?

PR #150, opened 2019-08-20
minor fix to keep text under mini-search from wrapping

erickpeirson:
Before:

image

After:

image


erickpeirson:
I tried ``white-space: nowrap`` at first, but it messes with flex, so went this route instead.

mhl10:
+9,126 −126  -- that's a massive diff! Looks like there are a lots of backtrace lines. Problem during sass compilation?

erickpeirson:
Argh what is that! Sorry, lemme table this for a sec and will investigate

PR #152, opened 2019-09-11
Fixing mistakes in definitions of categories.

bdc34:
Fixing mistakes in definitions of categories.
Not sure how I put the previous values in there.
ARXIVNG-2456


bdc34:
One thing is that stat.OT is not marked as general in classic. So it is not marked as general here either.

PR #154, opened 2019-10-16
ARXIVOPS-927: Update cs.LG category description

mhl10:


mhl10:
Merging per discussion with @jimentwood

PR #157, opened 2019-10-18
Pre-release merge for base v0.16.3

mhl10:
Includes updates from:
* ARXIVOPS-927 - update to cs.LG category description
* ARXIVNG-1543
* ARXIVNG-2745

Some additional minor commits may be forthcoming.


Pipfile
```diff
@@ -24,8 +24,9 @@ pylint = "<2"
 hypothesis = "*"
 sphinx = "*"
 sphinxcontrib-websupport = "*"
-sphinx-autodoc-typehints = "*"
+sphinx-autodoc-typehints = "==1.8.0"
 astroid = "<2.2"
+mypy-extensions = "==0.4.2"
```
mhl10: It's unclear to me how this and the corresponding hash in `Pipfile.lock` got in a state causing the builds to fail, but seems fixed now...

PR #158, opened 2019-10-29
ARXIVNG-2757 accessible links

SBBCornell:
added bolding to links to meet accessibility criteria (sass changes only)


SBBCornell:
The accessibility branch of browse is dependent on this branch. Browse should not be deployed without the new version of base.

PR #167, opened 2020-01-27
adjusts base colors to accessible values

SBBCornell:
The color palette in base wasnt fully compliant with WCAG contrast guidelines. This PR fixes that. See image for a visual comparison of changes. Note that the reds have not changed, blue only a smidge, while orange and green needed to shift quite a bit.

accessible-color-changes

This PR also includes some edits to a readme file, and I copied a group of styles from submission-ui that would be better hosted in base so that we can use them other places. But the main thing for review is the colors.


SBBCornell:
> Side comment: I was wondering if there exists is a tool that you may feed a css to and it analyzes all colors in terms of accessibility.

I just found a tool that I used on this ticket that lets you bulk upload a list of hex colors. It will import them as swatches and tell you the contrast ratio to your background. Here is my final [palette of greens](https://leonardocolor.io/?colorKeys=%23008312%2C%2300a51a%2C%23eaf8ec&base=f1f8f2&ratios=-1%2C3.5%2C4.56%2C6.51%2C8&mode=HSLuv) for example. And we also use Siteimprove for analyzing rendered pages, per CU's recommendation. Siteimprove flags colors as well as other criteria failures.

PR #168, opened 2020-02-05
ARXIVNG-2959: make analytics configurable

mhl10:
This adds analytics configuration options and also moves the analytics trackers into macros.


mhl10:
I will try to figure out why the tests are failing before I merge this PR.

PR #170, opened 2020-02-06
Moves alerts to macros to allow control of them in submission-ui

bdc34:
"There are are error messages originating from arxiv-base that are displaying above the pagination. We should consolidate error display in the submit-ui templates to below the page title."
https://arxiv-org.atlassian.net/browse/ARXIVNG-2917

### DIFF
--- a/arxiv/base/templates/base/base.html
+++ b/arxiv/base/templates/base/base.html
@@ -2,39 +2,23 @@
 <html lang="en">
   <head>
     {% include "base/head.html" %}
-    <!-- Custom style sheets or other head includes -->
-    {% block addl_head %}
+    {% block addl_head %}{# Custom style sheets or other head includes #}
     {% endblock addl_head %}
   </head>
   <body>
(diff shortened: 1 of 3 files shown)


arxiv/base/templates/base/base.html
```diff
@@ -18,26 +17,11 @@
...
-              {% if message.dismissable %}<button class="delete notification-dismiss"></button>{% endif %}
-              {% if message.title %}<h2 class="is-size-5 is-marginless">{{ message.title }}</h2>{%- endif %}
-              <p>{{ message.message }}</p>
-            </div>
-          {%- endfor %}
-        </div>
-      {%- endif %}
-    {%- endwith %}
-
+    {%block alerts %}{{ macros.alerts(get_alerts()) }}{% endblock alerts %}
```
mhl10:
```suggestion
    {% block alerts %}{{ macros.alerts(get_alerts()) }}{% endblock alerts %}
```

PR #171, opened 2020-02-11
Adds dist_version to arxiv.base.config

bdc34:
This uses dist_version get_version in arxiv.base.config for APP_VERSION so the static assets are severed from a version based on the same version in the git tag.


bdc34:
I noticed that this includes some commits that were squashed. I'll re-do this

bdc34:
Actually, this looks like the correct PR.

PR #173, opened 2020-02-11
Travis testing fixes

bdc34:
These are some changes that will hopefully fix the strange behavior we are seeing in travis.

I'll squash these since I was making a bunch of commits to test travis.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -8,15 +8,10 @@ os:
 python:
 - '3.6'
 install:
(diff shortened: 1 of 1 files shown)


mhl10:
Not sure why mypy tests are now failing, but looks good.

bdc34:
> Not sure why mypy tests are now failing, but looks good.

I'm not sure either.

PR #174, opened 2020-02-13
Changes to tagging check

bdc34:
This changes the tagging check.

1. It no longer checks for regressive tags. If tag 1.3 exists it will let you make 1.2.
2. It no longer sys.exit(1) on a bad version. Bad versions will not get into pypi because pypi will reject the upload.

This mainly will build the version.py file when there is a good tag.


mhl10:
I pushed a couple minor mypy-related fixes do your branch

PR #192, opened 2020-08-13
Adding some docs to docker_build_and_push

bdc34:


bdc34:
The build scripts deploy/push_base_image.sh and deploy/push_fourohfour_image.sh were deleted because the are use the generic docker_build_and_pusy.sh.

bmaltzan (approved):

PR #193, opened 2020-09-23
Changes abs page DOI link to handle more cases.

bdc34:
The DOI link in the abs metadata now accepts a very broad range
of values to trun into DOI links.

DOI links in other areas like the abstract are unaffected.

ARXIVNG-3423


bmaltzan (approved):
Looks good.

PR #195, opened 2020-10-16
Pre-release merge for arxiv-base 0.16.9rc2

mhl10:


SBBCornell (approved):
looks good!

bdc34 (changes requested):
Does SASS need to be built to CSS?

mhl10:
> Does SASS need to be built to CSS?

@SBBCornell, Brian is referring to the changes in arxiv/base/static/sass/_custom.sass

SBBCornell:
> > Does SASS need to be built to CSS?
> 
> @SBBCornell, Brian is referring to the changes in arxiv/base/static/sass/_custom.sass

@mhl10 @bdc34 I double checked the sass and css files and the change from _custom.sass is in arxivstyles.css, so that step did happen already.

bdc34 (approved):

PR #207, opened 2021-02-17
phoenix framework test of bootstrap

SBBCornell:
This branch is for testing out Bootstrap as a framework for future use on Phoenix. If running it locally you can see all changes on this page: http://127.0.0.1:5000/testing-bootstrap

I'll be building and testing a 'lego blocks' framework next. It might be most efficient to wait until both are ready for review.

Here is my analysis based on the success criteria defined here https://confluence.cornell.edu/display/arXiv/Decision%3A+a+Phoenix+UI+framework

1) is it useful to the whole team? 
- It is quite easy to work with to build layouts quickly. I actually don't remember the purpose of this criteria and would like other's input. 

2) is it easy to maintain over time?
- We will be able to keep all bootstrap source files separate from our customizations, which will make it straightforward to find and edit styles. 
- Discussions around how to rapidly deploy styles across all nodes, and manage versions, are ongoing and will also be important to meeting this criteria. 
- Ofcourse major bootstrap updates in the future (ones that substantially change style names or how they are used) would take quite a bit of work to implement across the site.

3) is it minimally-opinionated?
- It follows (and sets) expected industry standards with their grid and class system. It uses a ton of class names, which I dont love, and offers way more options than we will use. But its to be expected with a super-flexible system like this that wants to meet many needs, not just ours. 

4 and 5) does it follow a11y best practices for assistive technology and diverse audiences? All components pass WCAG Level AA criteria (papers will be exempted for now)?
- The form elements are doing better on accessibility than expected. Though most elements needed some small changes to pass Cornell's a11y criteria (WCAG 2.0 AA), no major edits were needed.
- There is an a11y issue with tooltips that would take some kind of hack to the tooltip js, probably not a big deal. The tooltip html is generated outside the <body> element in the dom, and Cornell's accessibility testing tool doesn't like that.
- I would give them a solid 'OK' on using semantic HTML elements when possible. I havn't reviewed every component but some use divs when semantic elements could be used instead. There is an overall abundance of nested divs.

6) can we self host with no unnecessary external dependencies?
- bootstrap 5 dropped jquery. I guess Bootstrap itself would be the dependency, but we can self-host all of the source files and it is fully open source.

7) Does it use SCSS, valid HTML5, and vanilla JS?
- Yes

8) Does it minimize use of JS, use CSS-only solutions when available and comparable?
- bootstrap uses javascript for many of their components. There are alternatives, like tabs can be built around radio buttons without js. But I dont know how much of an issue this is as long as the solution is quite accessible. I think we will need to build the lego blocks framework first and compare.

9) is it a white label solution, and can we limit references to other entities or orgs?
- Because we can self host we can rename files and make sure the name bootstrap doesnt appear. Someone who is familiar with bootstrap would still be able to tell because of their distinctive html and class structure, but it doesnt shout it loudly. 

10) Is it fast for end users?
- It does not seem slow. I think we'll need to compare to the lego blocks solution.


SBBCornell:
Dev team decision: to go with bootstrap5 and stop development of the custom lego blocks test.

Next I will update the bootstrap5 implementation from CDN to self-hosted, be a little choosy with our implementation so were only pulling in what we will reasonably use, and further cleanup the main template files.

SBBCornell:
I think this branch is ready to be pushed to Phoenix and used to format browse. I might need a walk through on those steps the first time through.

Also, I mainly worked in /templates and /static. I would love input from others on whether there are other files or directories that can be simplified or are unnecessary.

bdc34 (approved):
This seems fine to me. I think this could be simplified by moving it to its own GH repo. The templates, macros and assets could all be in some very simple repo with very little python code. This would make updating repos like arxiv-base to use a new version of arxiv-the-looks would very simple.

bdc34:
The build on this is failing but if I follow the link to the travis page I see an error related to the taxonomy/definitions.py file with the text "that illustrate the application of mathematics to"

I do not find that text in that file on line 1206 or at all. No changes were made to that file in this branch. I can run the tests locally with no errors.

I suspect travis is having some sort of problem.

mhl10:
I'm having some trouble viewing the `/phoenix` test endpoint when running the flask app:

Screen Shot 2021-03-03 at 11 26 45 AM

mhl10:
> I'm having some trouble viewing the `/phoenix` test endpoint when running the flask app:
> 
> Screen Shot 2021-03-03 at 11 26 45 AM

Looks like you already fixed that, thanks!

mhl10 (approved):
Looks good, thank you for the quick fix!

bmaltzan (approved):
This is great!

bmaltzan:
Looks like one of the builds is failing when trying to generate a package version for arxiv-base:
`git describe --dirty
fatal: No names found, cannot describe anything.
The command "git describe --dirty" failed and exited with 128 during .`

bmaltzan:
"--always" seems to have worked, generating a short sha reference as the build version.
Also increased the clone depth, so the format is consistent again and more unique.
https://docs.travis-ci.com/user/customizing-the-build#git-clone-depth

bdc34:
I don't think we are going to use this any time soon. We can leave the branch around.

PR #215, opened 2022-01-13
updates acknowledgement url

SBBCornell:
Minor change that updates the acknowledgement url in the browse header.


afromme (approved):
Thank you!

PR #218, opened 2022-08-10
Bump to fix depbot issues

bdc34:
This should close several of the depbot issues.

It is working toward closing dep issues in arixv-auth.

ARXIVNG-5128


arxiv/forms/csrf.py
```diff
@@ -116,7 +111,12 @@ def validate_csrf_token(self, form: 'CSRFForm', field: Field) -> None:
         """Validate the CSRF token passed with form data."""
         if field.data is None:
             raise ValidationError('Missing CSRF token')
-        digest, expires = self._split(field.data)
+
+        parts = field.data.split('::',1)
+        if len(parts) != 2:
+            raise ValidationError('CSRF token is malformed')
+
+        digest, expires = parts
```
DavidLFielding, line 119: I'm curious, is this new check due to observed corruption in the token?

arxiv/release/README.md
```diff
@@ -1,3 +1,22 @@
+# How to make a new version arxiv-base
+To make a new version of arxiv-base that gets uploaded to PyPI, commit
+a tag that is a valid python version in git and travis should upload
+the package to PyPI. Other repos can be set up in a similar manner.
```
DavidLFielding, line 4: The above makes it sound like the tag commit triggers an automatic upload to PyPi. The documentation below looks like the manual steps I'm used to following when uploading to PyPi.

DavidLFielding (approved):
Looks good. Only a few minor comments.

arxiv/release/README.md
bdc34: I think I'm going to get rid of most of this stuff. I wanted to automate it and it was working but then it didn't get used.

arxiv/forms/csrf.py
bdc34: I'm not sure how this wasn't caught before since it was making a unit test fail.

PR #219, opened 2022-08-23
Moves from python 3.6 to 3.10

bdc34:
Moved to poetry instead of pipenv. They are about the same but poetry has build and upload commands.


arxiv/base/logging.py
```diff
@@ -16,26 +16,9 @@
 from datetime import datetime
 from pytz import timezone, utc
 
-from flask import request
-
 from arxiv.base.globals import get_application_config
 
 
-class RequestFormatter(logging.Formatter):
```
bdc34, line 24: Removing this since it is not used. Simplified to just a normal logger formatter.

arxiv/base/tests/test_logging.py
```diff
@@ -23,56 +21,3 @@ def test_get_logger_no_app_nor_request(self):
         stream.close()
         self.assertIn('ERROR: "foo"', captured_value,
                       "Should log normally even if request is not present")
-
```
bdc34, line 26: Tests related to removed logger features removed.

arxiv/forms/csrf.py
```diff
@@ -1,65 +1,14 @@
-"""
```
bdc34, line 1: Removing CSRF protection. 

This is only in use the the unused code from the submission-ui repo.
 
WTForms has instructions for configuring CSRF.  We can use that instead and it is easy to setup.

PR #224, opened 2023-04-18
ARXIVCE-330: add .xyz domains to the bleach tlds

bmaltzan:
https://arxiv-org.atlassian.net/browse/ARXIVCE-330

### DIFF
--- a/arxiv/base/urls/links.py
+++ b/arxiv/base/urls/links.py
@@ -113,7 +113,7 @@ IP_ADDRESS = (
 )
 """Regex to match an IP address."""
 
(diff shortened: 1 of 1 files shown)


DavidLFielding (approved):

jonathanhyoung (approved):
Would upgrading bleach (or sending them a PR) be a better solution?

jweiskoff:
> Would upgrading bleach (or sending them a PR) be a better solution?

see: https://arxiv-org.atlassian.net/browse/ARXIVCE-330?focusedCommentId=272279

PR #240, opened 2024-03-11
Arxivce 1273 browse refactor

mnazzaro:


arxiv/base/config.py
```diff
@@ -156,3 +156,12 @@
...
 """Analytics tracker site ID."""
+
+TRACKBACK_SECRET = os.environ.get("TRACKBACK_SECRET", "baz")
+
+DEFAULT_DB = "sqlite:///../tests/data/browse.db"
+DEFAULT_LATEXML_DB = "sqlite:///..test/data/latexml.db"
+
+CLASSIC_DB_URI = os.environ.get("CLASSIC_DB_URI", DEFAULT_DB)
+LATEXML_DB_URI = os.environ.get("LATEXML_DB_URI", DEFAULT_LATEXML_DB)
+ECHO_SQL = os.environ.get("ECHO_SQL", True)
```
kyokukou, line 167: not sure we would want the default to be true

arxiv/config/__init__.py
```diff
@@ -0,0 +1,167 @@
...
+
+EXTERNAL_URL_SCHEME = os.environ.get("EXTERNAL_URL_SCHEME", "https")
+BASE_SERVER = os.environ.get("BASE_SERVER", "arxiv.org")
+
+AUTH_SERVER = os.environ.get("AUTH_SERVER", BASE_SERVER)
+"""Hostname for auth paths like /login and /logout.
+
+Usually the same as BASE_SERVER but can be configured.
+"""
+
```
kyokukou, line 23: for completeness RSS also has a server (although I doubt anyone outside of RSS cares)

arxiv/db/models.py
```diff
@@ -0,0 +1,2009 @@
...
+
+class BogusCountry(Base):
+    __tablename__ = 'arXiv_bogus_countries'
+
+    user_id: Mapped[int] = mapped_column(Integer, primary_key=True, server_default=FetchedValue())
+    country_name: Mapped[str] = mapped_column(String(255), nullable=False, server_default=FetchedValue())
+
+
+
+t_arXiv_bogus_subject_class = Table(
```
kyokukou, line 285: why are some tables and some classes for the models?

arxiv/document/metadata.py
```diff
@@ -0,0 +1,313 @@
...
+    __source_format_options__ = Literal['tex', 'pdftex', 'ps', 'html', ]
+    source_format: Optional[__source_format_options__] = None
+
+    version: int = 1
+    """Version of this article."""
+
+    license: License = field(default_factory=License)
+    """License associated with the article."""
+
+    version_history: List[VersionEntry] = field(default_factory=list)
```
kyokukou, line 116: im not sure if we want version history to be a part of all our DocMetadata objects, version history isnt a column in the doc metadata table and would require additional fetching of all version entries for a document when they may not be needed/wanted
Maybe whetever relies on version history could instead take a List of DocMetadata objects?

arxiv/base/config.py
mnazzaro: Fair point

arxiv/config/__init__.py
```diff
@@ -0,0 +1,167 @@
...
+]
+"""
+URLs for external services, for use with :func:`flask.url_for`.
+
+For details, see :mod:`arxiv.base.urls`.
+"""
+
+# In order to provide something close to the config_url behavior, this will
+# look for ARXIV_{endpoint}_URL variables in the environ, and update `URLS`
+# accordingly.
```
mnazzaro, line 103: Can I just delete this? Seems like an annoying bug waiting to happen

arxiv/identifier/iteration.py
```diff
@@ -0,0 +1,45 @@
...
+    def __init__ (self,
+                  start_yymm: str,
+                  end_yymm: str,
+                  categories: Optional[List[Category]] = None,
+                  source_formats: Optional[List[SOURCE_FORMAT]] = None,
+                  only_latest_version: bool = False):
+        
+        with get_db() as session:
+            self.query = session.query(Metadata.paper_id, Metadata.version) \
+                .filter(Metadata.paper_id >= start_yymm) \
```
kyokukou, line 23: fyi this will only work for papers after march 2007

arxiv/identifier/iteration.py
```diff
@@ -0,0 +1,45 @@
...
+                .filter(Metadata.paper_id < f'{end_yymm}.999999')
+            if categories:
+                self.query = self.query.filter(or_(*[Metadata.abs_categories.like(f'%{category.id}%') for category in categories]))
```
kyokukou, line 26: possible hiccup here due to some entries not including the most modern/relevant name of their category. Technically this code is fine, but to find every paper whatever calls this would have to include all possible category names (both versions of aliases and subsumed arxivs)

arxiv/py.typed
kyokukou, line 1: adding empty file?

ntai-arxiv (changes requested):
Sorry about my not-so-curret docstring.

arxiv/identifier/iteration.py
mnazzaro: Hmm yeah I wrote this piece of code sort of carelessly, and I don't know what it would be useful for. I'd like to search/sort by announce date but that doesn't seem easy with the columns available to me. I may just delete it and write again when I start work on converting the full corpus to HTML

arxiv/document/metadata.py
bdc34: The VersionEntity is outside the ORM but if it were more directly supported we could specify loading techniques: https://docs.sqlalchemy.org/en/20/orm/queryguide/relationships.html

arxiv/config/__init__.py
bdc34: This does seem too auto-magical for me. Isn't there a list of URLs that can be configured in a sub-app?

arxiv/identifier/iteration.py
kyokukou: ive spent way too much time figuring how to search for all papers of a given demographic. LMK when you work on that

arxiv/document/metadata.py
```diff
@@ -0,0 +1,313 @@
...
+from ..taxonomy import definitions
+from ..taxonomy.category import Category, Group, Archive
+from ..identifier import Identifier
+from ..license import License
+from .version import VersionEntry
+
+from ..base.urls import canonical_url
+
+
+@dataclass(frozen=True)
```
bdc34, line 16: I've found frozen has been unpleasant here because prevents the creation of subclasses. 

No change needed now.

arxiv/document/parse_abs.py
```diff
@@ -0,0 +1,309 @@
...
+    alt_comp.append(abstract)
+    alt_comp.append('')
+    return alt_comp
+
+
+def _get_tz() -> ZoneInfo:
+    """Gets the timezone from the flask current_app."""
+    global _fs_tz
+    if _fs_tz is None:
+        _fs_tz = ZoneInfo(current_app.config["FS_TZ"])
```
bdc34, line 307: The use of `current_app` ties this code to being run under flask.

arxiv/ops/fastly_log_ingest/app.py
```diff
@@ -61,7 +61,10 @@ def _status_to_severity(self, status: int):
     def __init__(self, message: Message, arrival_time: float):
         self.message = message
         self.arrival_time = arrival_time
-        self.data: dict = json.loads(message.data.decode('utf-8').strip())
+        try:
+            self.data: dict = json.loads(message.data.decode('utf-8').strip())
+        except:
+            logger.warning(f'Failed to log the following message: {message.data.decode("utf-8")}')
```
bdc34, line 67: Since the decode could have been the problem, use decode('utf8', errors='replace') in the error message.

Please log the exception message too.

arxiv/urls/__init__.py
```diff
@@ -0,0 +1,12 @@
+from ..identifier import Identifier
+
+def pdf_canonical_url (identifier: Identifier) -> str:
+    # This won't include the version unless identifier 
+    # was explicitly initialized with a version
+    return f'https://arxiv.org/pdf/{identifier.idv}'
```
bdc34, line 6: Not sure where this is used but might be good to use config.CANONICAL_SERVER

But maybe some other time.

arxiv/util/tests/test_authors.py
```diff
@@ -1,7 +1,7 @@
 """Tests for author and affiliation parsing."""
 from unittest import TestCase
 
-from ..authors import parse_author_affil, split_authors
```
bdc34, line 4: Relative imports are often used in tests so if the arxiv.util is moved, the tests don't need to change. This is probably not an issue with modern refactoring IDEs.

No change requested.

mypy.ini
```diff
@@ -1,5 +1,5 @@
 [mypy]
-mypy_path = $MYPYPATH:./sqlalchemy-stubs
+# mypy_path = $MYPYPATH:./sqlalchemy-stubs
```
bdc34, line 2: Please remove if not in use.

pyproject.toml
```diff
@@ -4,19 +4,19 @@ version = "1.0.1"
...
 ]
    
 [tool.poetry.dependencies]
-python = "^3.10"
-
-flask = "~2.2.5"
-Werkzeug = "~2.3.8"
+python = "^3.11"
+pydantic = "==1.10.*"
+flask = "3.0.2"
```
bdc34, line 19: A less constrained version for flask like 3.0.* might be good.

bdc34 (changes requested):
Looks good. Some minor changes requested. Mostly I've left comments.

bdc34 (approved):

kyokukou (changes requested):
This currently has a bug in it with the new category structure, I've fixed it in a branch that hasn't been pulled into this one yet. Also I think it would be good to get all of the tests running.

PR here: https://github.com/arXiv/arxiv-base/pull/244

ntai-arxiv (approved):

kyokukou (approved):

PR #241, opened 2024-03-20
Combining tcategory and category

kyokukou:
Redesigned the Category, Archive, and Group classes to have properties and methods matching what we expect them to contain rather than just being a string. Combined the tcategory typed dictionaries into these. Rewrote the definitions in this class form instead of a dictionary with keys form.


arxiv/base/templates/base/macros.html
```diff
@@ -261,7 +261,7 @@ <h1 class="title mathjax"><span class="descriptor">Title:</span>{{ title|tex2utf
 
 Title: {{ title }}
 Authors: {{ authors }}
-Categories: {{ primary_category }}{%- for category in secondary_categories|sort(attribute='id') -%} {{ category }}{%- endfor %}
+Categories: {{ primary_category.id }}{%- for category in secondary_categories|sort(attribute='id') -%} {{ category }}{%- endfor %}
```
bdc34, line 264: I wonder if we should define `__str__` for Category so this could be `{{primary_category}}` ?

arxiv/document/metadata.py
```diff
@@ -161,14 +161,14 @@ def get_browse_context_list(self) -> List[str]:
         if self.primary_category:
             options = {
                 self.primary_category.id: True,
-                definitions.CATEGORIES[self.primary_category.id]['in_archive']: True
+                self.primary_category.in_archive: True
```
bdc34, line 164: I like this improvement.

arxiv/taxonomy/category.py
```diff
@@ -1,73 +1,90 @@
...
-class Group(Category):
-    """Represents an arXiv group--the highest (most general) taxonomy level."""
+    in_archive: str
+    is_general: bool
+    description: Optional[str]
 
-    lookup = GROUPS
+    def get_archive(self) -> Archive:
+        """Returns parent archive."""
+        from .definitions import ARCHIVES
```
mnazzaro, line 89: Import at the top unless there's a good reason not to

arxiv/document/metadata.py
```diff
@@ -296,7 +292,7 @@ def raw(self) -> str:
             rv += f"Comments: {self.comments}\n"
         # skipping proxy to avoid harvesting of email addresses
         if self.report_num:
-            rv += "Report-no: {self.report_num}\n"
+            rv += f"Report-no: {self.report_num}\n"
```
bdc34, line 295: :+1:

mnazzaro (changes requested):
Looks really good. Just one comment that may not be warranted

bdc34 (approved):

arxiv/base/templates/base/macros.html
kyokukou: I'm not sure it has a couple string methods - notably display() which creates a longer string meant for display purposes. There is also full_name

arxiv/taxonomy/category.py
kyokukou: its circular if you do it at the top :/ ARCHIVES relies on the structure in taxonomy to define everything. But I still needed access to ARCHIVES within some of the functions the class uses

mnazzaro: Oof yeah. Since they are codependent, it may make sense to put them in the same file. I think the taxonomy dir should have a __init__.py too?

mnazzaro: Alternatively, those methods can be module methods instead of class methods like `def get_archive (group: Group) -> Archive:`

kyokukou: As far as putting them in the same file, there's still an issue with what order you define them in. And honestly the definitions file is huge and I wanted the category file with type definitions to be more readable/usable

mnazzaro (approved):

PR #255, opened 2024-04-22
Tiny GCP library for the service account to get the identity token

ntai-arxiv:


gcp/service-auth/pyproject.toml
```diff
@@ -0,0 +1,25 @@
...
+authors = ["arxiv.org"]
+license = "BSD-3 License"
+readme = "README.md"
+packages = [
+    { include = "gcp_service_auth" }
+]
+include = ["README.md"]
+
+[tool.poetry.dependencies]
+python = "^3.8"
```
mnazzaro, line 14: Can we make it work with python 3.11?

gcp/service-auth/tests/test_get_token.py
```diff
@@ -0,0 +1,39 @@
...
+from typing import Generator
+
+import pytest
+import logging
+import datetime
+from pathlib import Path
+
+from gcp_service_auth import GcpIdentityToken
+
+dev_target = "https://gcp-genpdf-6lhtms3oua-uc.a.run.app"
```
mnazzaro, line 13: I think this should be a config/env variable

gcp/service-auth/tests/test_get_token.py
```diff
@@ -0,0 +1,39 @@
...
+
+@pytest.fixture(scope="module")
+#@pytest.mark.with_op
+def gcp_browse_cred() -> Generator[str, str, str]:
+    logging.basicConfig(level=logging.DEBUG)
+    cred_file = os.path.join(Path(__file__).parent, "browse-local.json")
+    cred_file = cred_file
+    if not os.path.exists(cred_file):
+        subprocess.run(["op", "document", "get", "4feibaz4tzn6iwk5c7ggvb7xwi", "-o", cred_file])
```
mnazzaro, line 22: What is this random alphanumeric string?

mnazzaro (approved):
Looks good to me. Small changes maybe but not super important

gcp/service-auth/pyproject.toml
ntai-arxiv: CIT sycn-node still uses 3.8, and wanted to NOT install python 3.10 on it if I could.

gcp/service-auth/tests/test_get_token.py
ntai-arxiv: Will take it out. My bad.

gcp/service-auth/pyproject.toml
ntai-arxiv: > Can we make it work with python 3.11?

It would work any version above 3.8.

PR #256, opened 2024-04-23
Tiny change to HTML key pattern to allow HTML conversions without the v

mnazzaro:


bdc34 (approved):

ntai-arxiv (approved):
Why the extra '/' on one and not for the other?

PR #263, opened 2024-04-30
Swap order of acm-class and msc-class to conform to schema.

cbf66:
How do we test this?


jweiskoff:
The reported oai url that showed the error was in the ticket: https://arxiv-org.atlassian.net/browse/ARXIVCE-1659 and is: 

https://export.arxiv.org/oai2?verb=GetRecord&identifier=oai:arXiv.org:2305.04323&metadataPrefix=arXivRaw

bmaltzan (approved):

jweiskoff (approved):

PR #264, opened 2024-05-01
Add env vars for setting transaction isolation level for both dbs

mnazzaro:


cbf66 (commented):
Can this be a per-connection parameter please? Maybe with these env vars providing the default value?

jonathanhyoung (commented):
This is what I suggested, but it isn't what we want, @mnazzaro .  We want to be able to specify the transaction isolation level *per request* (or per transaction).

jonathanhyoung (approved):

PR #274, opened 2024-05-17
Handle error for gcs list blob

mnazzaro:


arxiv/files/object_store.py
```diff
@@ -127,7 +127,10 @@ def list(self, prefix: str) -> Iterator[FileObj]:
         'ps_cache/arxiv/pdf/1212/1212.12345' or
         'ftp/cs/papers/0012/0012007'.
         """
-        return self.bucket.client.list_blobs(self.bucket, prefix=prefix)  # type: ignore
+        try:
+            return self.bucket.client.list_blobs(self.bucket, prefix=prefix)  # type: ignore
+        except:
```
bdc34, line 132: This will hide too many errors. I'd much rather have the user see a 500 error than a response built with an incorrectly empty list from this.

Replace this except with something like

    except Exception as inner_ex:
        raise RuntimeException(f"Could not get obj_store.list() for {bucket} {prefix}") from inner_ex

bdc34 (approved):

bdc34 (approved):

PR #276, opened 2024-05-20
(1) Not select the with_op by default as you may not have 1password CLI

ntai-arxiv:
(2) Make the test to run from the arxiv-base directory

### DIFF
--- a/gcp/service_auth/tests/test_get_token.py
+++ b/gcp/service_auth/tests/test_get_token.py
@@ -1,4 +1,5 @@
 import os
+import sys
 import subprocess
 import time
 from typing import Generator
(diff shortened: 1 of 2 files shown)


gcp/service_auth/tests/test_get_token.py
```diff
@@ -7,33 +8,47 @@
...
+    from gcp_service_auth import GcpIdentityToken
+    pass
+                                 
 
-TEST_URL = os.environ.get('TEST_URL')
 
 @pytest.fixture(scope="module")
 #@pytest.mark.with_op
-def gcp_browse_cred() -> Generator[str, str, str]:
+def gcp_browse_cred() -> (Generator[str, str, str], str):
```
jonathanhyoung, line 24: I'm confused.  You yield (cred_file, test_url), then return "". Neither matches the signature.  

You probably want the return type to be Generator[str, str] and not have the final return "" .

gcp/service_auth/tests/test_get_token.py
```diff
@@ -7,33 +8,47 @@
...
+    op = subprocess.run(["op", "item", "get", "4feibaz4tzn6iwk5c7ggvb7xwi", "--format", "json"], stdout=subprocess.PIPE)
+    test_meta = json.loads(op.stdout)
+    test_url = ""
+    for field in test_meta['fields']:
+        if field["id"] == "heltf7phky3h6rnlvd7zi3542u":
+            test_url = field["value"]
+            break
+    yield (cred_file, test_url)
     os.remove(cred_file)
     return ""
```
jonathanhyoung, line 40: Not your code, but IMHO "file" and "filename" are well-understood, distinct data types.  I would call this "cred_filename" not "cred_file".  

What is the purpose of assigning cred_file = cred_file ?

gcp/service_auth/tests/test_get_token.py
```diff
@@ -7,33 +8,47 @@
...
 
 @pytest.mark.with_op
-def test_get_token(gcp_browse_cred: str) -> None:
-    os.environ["GOOGLE_APPLICATION_CREDENTIALS"] = gcp_browse_cred
+def test_get_token(gcp_browse_cred: (str, str)) -> None:
+    os.environ["GOOGLE_APPLICATION_CREDENTIALS"] = gcp_browse_cred[0]
```
jonathanhyoung, line 44: Bizarre. gcp_browse_cred is a generator, not an array?!?

jonathanhyoung (changes requested):

gcp/service_auth/tests/test_get_token.py
jonathanhyoung: What did the magic strings 4feibaz4tzn6iwk5c7ggvb7xwi and heltf7phky3h6rnlvd7zi3542u come from?

gcp/service_auth/tests/test_get_token.py
jonathanhyoung: Is this somehow calling gcp_browse_cred (the generator) and getting one tuple of credentials?  wow. I hate magic like this .

gcp/service_auth/tests/test_get_token.py
ntai-arxiv: TRUE. My brain was working for yield.

gcp/service_auth/tests/test_get_token.py
ntai-arxiv: It is the output from 1password CLI "op". The cred in 1p has these keys.

gcp/service_auth/tests/test_get_token.py
ntai-arxiv: pytest fixture is hacky. it is what it is.

bdc34 (commented):
Why are we making it so our tests require the onepassword cli tool?

jonathanhyoung (approved):

ntai-arxiv:
> Why are we making it so our tests require the onepassword cli tool?

If you run the pytest from arxiv-base root, it is turned off. OTOH, if you need to test the token generation, it needs a credential to run it against GCP. I stored it in 1p.

Do you have a suggestion for it? @bdc34

bmaltzan:
> > Why are we making it so our tests require the onepassword cli tool?
> 
> If you run the pytest from arxiv-base root, it is turned off. OTOH, if you need to test the token generation, it needs a credential to run it against GCP. I stored it in 1p.

Hmm, can I trust automated tests with my 1p and gcloud credentials?

ntai-arxiv:
> > > Why are we making it so our tests require the onepassword cli tool?
> > 
> > 
> > If you run the pytest from arxiv-base root, it is turned off. OTOH, if you need to test the token generation, it needs a credential to run it against GCP. I stored it in 1p.
> 
> Hmm, can I trust automated tests with my 1p and gcloud credentials?

@bmaltzan The test is not meant to run in the automated test. You need to run it on your local machine with 1password CLI installed. The test has the marker "with_op" for it.

PR #288, opened 2024-06-25
purge all for paper, default to hard purges

kyokukou:


arxiv/integration/tests/test_fastly.py
```diff
@@ -32,8 +48,7 @@ def test_purge_multiple_keys(self, MockApiClient, MockPurgeApi: PurgeApi):
 
         mock_api_instance.bulk_purge_tag.assert_called_once_with(
             service_id="umpGzwE2hXfa2aRXsOQXZ4",
```
bmaltzan, line 50: It's fine, but better to define constants.
Not sure why you have two.

bmaltzan (approved):

PR #297, opened 2024-07-22
ARXIVCE-1264: group by ip, when counting pdf downloads

bmaltzan:
Create a table subquery to filter out partial-content pdf downloads, then group by paper id.

### DIFF
--- a/arxiv/ops/arxiv_stats/parse_paperid_remoteaddr_from_weblogs.sh
+++ b/arxiv/ops/arxiv_stats/parse_paperid_remoteaddr_from_weblogs.sh
@@ -2,30 +2,37 @@
 
 # File: parse_paperid_remoteaddr_from_weblogs.sh
 # Goal: returns the stats day by day, split out by categories
-# Desc: This script extracts paper_id:remote_addr pairs, from fastly web logs,
-#       counting downloads per ip, once per hour.
+# Desc: This script extracts paper_id:remote_addr pairs from fastly web logs.
+#       The inner query groups by ip, to avoid overcounting partial-content downloads.
+#       The outer query then groups/counts paper id by country.
 #
 # This script's query, is the checked in version, of what is scheduled here:
 # https://console.cloud.google.com/bigquery/scheduled-queries?project=arxiv-production
 #
-# It runs hourly, and writes here, when it completes:
+# The scheduled query runs hourly, stores the output here:
 #   projects/arxiv-production/topics/papers-downloaded-by-ip-recently
-# The sub:
+#
+# The scheduled query notifies on completion by pub/sub:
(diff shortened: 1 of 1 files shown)


arxiv/ops/arxiv_stats/parse_paperid_remoteaddr_from_weblogs.sh
```diff
@@ -2,36 +2,45 @@
...
-      REGEXP_EXTRACT(STRING(json_payload.path), r"^/pdf/([a-zA-Z\.-]*/?[0-9.]+)") as paper_id,
       STRING(json_payload.remote_addr) as remote_addr,
+      REGEXP_EXTRACT(STRING(json_payload.path), r"^/[^/]+/([a-zA-Z\.-]*/?[0-9.]+[0-9])") as paper_id,
+      STRING(json_payload.geo_country) as geo_country,
+      REPLACE(left(STRING(json_payload.path), 5), "/", "") as download_type,
       TIMESTAMP_TRUNC(TIMESTAMP_SUB(CURRENT_TIMESTAMP(), INTERVAL 3 HOUR), HOUR) as start_dttm
  FROM arxiv_logs._AllLogs
 WHERE log_id = "fastly_log_ingest"
-  AND left(STRING(json_payload.path), 4) = "/pdf"
+  AND REGEXP_CONTAINS(STRING(json_payload.path), "^/(html|pdf|src|e-print)/")
```
bdc34, line 39: /html could massively over count due to conference proceedings having paths like /html/{paperid}/figures/fig1.png /html/{paperid}/figures/fig2.png /html/{paperid}/figures/fig3.png etc

bmaltzan: The subquery counts 1 occurence of /html/{paperid}/ per hour, per ip.

bdc34 (changes requested):
I'm concerned about over counting conference proceedings due to the /html/ in the regex.

arxiv/ops/arxiv_stats/parse_paperid_remoteaddr_from_weblogs.sh
kyokukou, line 39: source and e-prints are the same and should be counted together

bmaltzan:
> I'm concerned about over counting conference proceedings due to the /html/ in the regex.

It's possible for Erin to also lookup the format, but I don't think we should.

I think the distinction of someone choosing to view teX generated html over choosing to view the pdf is interesting, and worthwhile for us to extract, but I think the primary thing people are looking at is the number of people reading the paper, so this is not misleading, if a little bit off.

arxiv/ops/arxiv_stats/parse_paperid_remoteaddr_from_weblogs.sh
bmaltzan: If so, why are we maintaining 2 urls?

PR #305, opened 2024-08-20
removes addition of donate text link

kyokukou:
its being replaced by a permanent button in browse


bdc34 (approved):

bdc34:
This will mean that non-browse app that use this template will lack the donate button.

kyokukou:
> This will mean that non-browse app that use this template will lack the donate button.

Is there something else we should do about that? Or is this another drop in the bucket of our static files system is a mess?

PR #307, opened 2024-08-22
Adds DOC_SERVER config for info.arxiv.org

bdc34:


bdc34:
These changes already exist in master.

PR #313, opened 2024-09-24
ARXIVCE-2389 new generic regex [a-z][a-z]+ 

jonathanhyoung:
This will match all latin-1 TLDs. Old code used a list from bleach.linkifier, which will always be out of date.

### DIFF
--- a/arxiv/base/urls/links.py
+++ b/arxiv/base/urls/links.py
@@ -31,6 +31,9 @@ handle FTP addresses.
 
 Updated 20 March, 2019: refactored to provide independent bleach attribute
 callbacks for each kind of link.
+
+Updated 24 Sep 2024: TLDs is just 2 or more letters.
+
 """
 from typing import (
     List,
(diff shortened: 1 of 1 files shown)


bmaltzan (approved):

bdc34 (approved):
This is worth a try.

arxiv/base/urls/links.py
```diff
@@ -110,7 +113,7 @@ def _without_group_names(pattern: Pattern) -> str:
 )
 """Regex to match an IP address."""
 
-TLDS = "|".join( list(set(bleach.linkifier.TLDS + ['xyz'])) )
+TLDS = "[a-z][a-z]+"     # TLDs are out of control. Just match on two or more alphabetic characters.
```
dginev, line 116: `[a-z]{2,24}` if you'd like to keep some defensive posture in place, which fits [today's entries](https://data.iana.org/TLD/tlds-alpha-by-domain.txt)

bmaltzan: Just noticing, that list contains numbers, uppercase, and hyphens: 
eg: XN--11B4C3D

dginev: Right... Unsure if arXiv wants to support the international/punycode TLDs.

I found this illustration useful:
https://stackoverflow.com/a/73155251

jonathanhyoung: The matching is case-insensitive (elsewhere).

I don't see the need to highlight the other URLs yet.

PR #328, opened 2024-10-18
Use Firefox instead of Chromium for Keycloak test

ntai-arxiv:
Chromium and its driver are snap based and it does not work with ACT. IOW, for testing, you don't want to rely on snapd up and running. I switched to use Firefox ESR since it is .beb based which in turn result in using the driver. 

Note that, this means the gecko driver is external and could become obsolete. When it happens, it may need to update the driver version which is unfortunate but the alternative is to be stuck with Chrome driver which is 20+ versions behind the stable Chrome stable release.


norbusan (approved):
LGTM

PR #336, opened 2024-11-06
Fixes some fastly_log_ingest problems

bdc34:
Adds `try/catch` to WorkItem loop
Adds README
Adds SIGTERM handling for quick finishing of queue on shutdown
Has both a PREFERRED_PER_BATCH and MAX_PER_BATCH
Better date handling
Fixes `notify()`

Before the changes this did about 200 messages per sec, after it does about 300.


norbusan (approved):
LGTM

PR #344, opened 2024-12-18
Make the cookie simpler. Completely ditch the access token

ntai-arxiv:
The access token can be in the authentication header.


jonathanhyoung (approved):
No attempt at backwards compatibility?

bdc34 (approved):

bdc34:
> No attempt at backwards compatibility?

I think this is the new keycloak JWT, not one of the existing formats.

ntai-arxiv:
> No attempt at backwards compatibility?

There is no old version outside of my computer and dev3, despite it is already version 5.

PR #357, opened 2025-03-04
ARXIVCE-825 code improvements

jonathanhyoung:
- Refactored code to offer context on certain errors
- Improved ALL CAPS "word" detection
- Improved HTML detection
- Added code to detect UTF-8 encoding issues
- Match word boundaries when matching "dr", "phd", etc.


arxiv/metadata/metacheck.py
```diff
@@ -59,10 +59,51 @@ def contains_outside_math(s1:str, s2:str) -> bool:
...
+
+class MetadataCheckReport:
+    def __init__(self):
+        self.disposition = OK
+        # map from complaint to context (may be "")
+        self.complaints = {}
+
+    def __repr__(self):
+        if self.disposition == OK:
+            return f"<MetadataCheckReport OK>"
```
norbusan, line 75: Human ruff: no f string if nothing is interpolated

arxiv/metadata/metacheck.py
```diff
@@ -59,10 +59,51 @@ def contains_outside_math(s1:str, s2:str) -> bool:
...
+        else:
+            return f"<MetadataCheckReport {self.disposition}: {self.get_complaints_strings()}>"
+        
+
+    def add_complaint(self, disposition, complaint_string:str, context:str = ""):
```
norbusan, line 80: human ruff: space between `:` and type, missing type for disposition

arxiv/metadata/metacheck.py
```diff
@@ -59,10 +59,51 @@ def contains_outside_math(s1:str, s2:str) -> bool:
...
+        self.disposition = combine_dispositions(self.disposition, disposition)
+        if complaint_string in self.complaints:
+            self.complaints[complaint_string].append(context)
+        else:
+            self.complaints[complaint_string] = [context]
+        #
```
norbusan, line 86: Could we use a `set` here to not do the inclusion checks?

arxiv/metadata/metacheck.py
```diff
@@ -103,15 +147,13 @@ def long_word_caps(s: str) -> bool:
...
-    * contain at least one capitalized letter (not digits or punctuation!), 
+    * is in all caps (TODO: strip trailing comma ","?)
     * and which are not in our list of KNOWN_WORDS_IN_ALL_CAPS.
     """
     num_matches = 0
     for word in s.split():
         if len(word) >= 6 and \
-           word.upper() == word and \
-           word.lower() != word and \
+           re.match( "^[A-Z]*$", word ) and \
```
norbusan, line 156: What are cases where `word.upper() == word` is not enough for this test?

arxiv/metadata/metacheck.py
```diff
@@ -120,14 +162,19 @@ def long_word_caps(s: str) -> bool:
...
-        if re.match( "[a-z][A-Z][a-zA-Z]*: [A-Z]", s):
+        if re.match( "[a-z][a-zA-Z-]*: [A-Z]", s):
+            return False
+        elif re.match( "[a-z][A-Z][A-Z]* ", s):
+            # Matches uGMS
+            return False
+        elif re.match( "[a-z]-[A-Z]", s):
             return False
         else:
             return True
```
norbusan, line 180: I don't understand that check. As far as I read that, titles like `Fixing the Incompleteness Theorem` would  be rejected?
- it contains a lowercase `re.match("[a-z]", s)`
- it does not start with a lower case
so it returns `True` ?
Am I missing something?

arxiv/metadata/metacheck.py
```diff
@@ -169,14 +216,14 @@ def language_is_not_english(s: str) -> bool:
         return False
```
norbusan, line 216: This comment is concerning `all_brackets_balanced` - do we reect titles with smilies, like
`The social reaction to ;-)`
?

arxiv/metadata/metacheck.py
dginev: > Am I missing something?

There are no starting hats `^` in the regexes. They match anywhere within the string.

arxiv/metadata/metacheck.py
```diff
@@ -195,179 +242,156 @@ def contains_html_elements(s: str) -> bool:
...
     #
     if contains_outside_math( r"\\textbf", v):
-        disposition = combine_dispositions(disposition, WARN)
-        complaints.append("Title: contains \\textbf")
+        report.add_complaint( WARN, "Title: contains \\textbf")
     #
     if contains_outside_math( r"\\texttt", v):
-        disposition = combine_dispositions(disposition, WARN)
-        complaints.append("Title: contains \\texttt")
+        report.add_complaint( WARN, "Title: contains \\texttt")
```
norbusan, line 308: All the `title contains \textXX` - are they actually bad? I used `\emph` etc at times.

arxiv/metadata/metacheck.py
norbusan: Ahhh ...re.match only matches at the beginning of the string ... TIL

arxiv/metadata/metacheck.py
jonathanhyoung: No. The first check is necessary in order to accumulate a list of contexts; I don't think I want a set. 

I can use a defaultdict(list) if you like.

arxiv/metadata/metacheck.py
jonathanhyoung: if word = "123.456", word.upper() == word , but that's not something this code wants to complain about.

arxiv/metadata/metacheck.py
jonathanhyoung: Yes, this code does reject smilies.

arxiv/metadata/metacheck.py
jonathanhyoung: I implemented the requirements I was given.

We get more cases of \emph in the abstracts; it's not clear what we want to support but we don't support all of TeX, just mathjax.

PR #387, opened 2026-02-02
Fix the utf-8 column type decorator not applied.

ntai-arxiv:
Sorry for the inconvenience. 

The column expression didn't know to ask the class to post process. 

The tests added (not sure this runs with GHA. I will find it out soon enough).

Testing started at 5:49 PM ...
Connected to pydev debugger (build 251.23774.444)
Launching pytest with arguments /home/ntai/arxiv/arxiv-base/tests/test_utf8_columns.py --no-header --no-summary -q in /home/ntai/arxiv/arxiv-base/tests

============================= test session starts ==============================
collecting ... collected 2 items

test_utf8_columns.py::TestTapirEmailTemplateUtf8::test_insert_and_read_utf8_data 
test_utf8_columns.py::TestTapirEmailTemplateUtf8::test_utf8_boundary_characters 

============================== 2 passed in 20.37s ==============================
PASSED [ 50%]PASSED [100%]
Process finished with exit code 0

https://github.com/sqlalchemy/sqlalchemy/blob/main/lib/sqlalchemy/sql/type_api.py#L1596-L1600

The intension of TypeDecorator is to allow the post-process the returned data, and not quite sure why it broke.  I swear that I've seen process_return_value was trigged in my debugger. It is now force-feeding the decorator type in the column experssion which I should not have to do it, as the column type has already been mapped in the column type.

### DIFF
--- a/arxiv/base/tests/test_docmeta.py
+++ b/arxiv/base/tests/test_docmeta.py
@@ -1,6 +1,5 @@
-
 from typing import Any, Dict
-from unittest import TestCase,mock
+from unittest import TestCase, mock
 from datetime import datetime
 
 from arxiv.document.metadata import DocMetadata, AuthorList, Submitter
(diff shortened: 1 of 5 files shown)


kyokukou (approved):
this worked when I tested it in modapi

PR #389, opened 2026-02-16
[WOMBAT-359] Adopt Utf8Text for arXiv_admin_log.logtext

ntai-arxiv:
- Utf8Strig, Utf8Text's column type definition is factored out as `column_factory`  The factory function takes care of mysql/sqlite3 column type generation. 
- db_codegen: Instead of relying on installing sqlacodegen, use the checked out code for convenience
- fixed mypy warnings of column_types.py
- Running github action testing skips the GCP access while running locally

### DIFF
--- a/arxiv/db/column_types.py
+++ b/arxiv/db/column_types.py
@@ -32,7 +32,7 @@ class BinaryStringComparator(UserDefinedType.Comparator):
             # Encode the string to bytes using UTF-8
             other = other.encode('utf-8')
         # Compare using BINARY cast on both sides
-        return operator(func.binary(self.expr), func.binary(other))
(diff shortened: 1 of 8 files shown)


arxiv/db/column_types.py
```diff
@@ -170,7 +169,7 @@ def bind_expression(self, bindvalue):
         """
         return func.binary(bindvalue)
```
bdc34: Change all uses of `func.binary()` with `func.cast(xyz, LargeBinary)` for wider compatiblility.

bdc34 (changes requested):
Add a test. This code needs to be tested.

bdc34:
Write a pytest that uses the `classic_db_engine` fixture and writes utf8 to a column that uses this and reads it back.

bdc34 (approved):

PR #393, opened 2026-02-18
[patch] legacy sessions in latest arxiv-base

dginev:
1. I was inspired by https://github.com/arXiv/arxiv-base/commit/e2440aafb64956eaac483255bf6147d74e9bfdb1 in fixing the relative import.

2. Allowing JWT_SECRET to be unset as a fallback (reusing from a non-JWT context)


bdc34 (approved):

dginev:
The testing mocks expect the relative setup quite precisely, so this may be too brave of a refactor. I will see if I can find a different workaround in the upstream service.

PR #394, opened 2026-02-23
Adds proxy and sword to legacy auth scope

bdc34:
Adds missing proxy and sword to scope provided by arxiv.auth.legacy


DavidLFielding (approved):
Looks good.

PR #395, opened 2026-02-25
QA-100 checksum metadata

bdc34:
Adds `checksum_metadata` and a test.

Moves a Protocol to __init__.py for sharing.

Removes \r from some files. This change is what is causing the large diffs that seem to be nothing. 

https://arxiv-org.atlassian.net/browse/QA-100

### DIFF
--- a/arxiv/metadata/tests/test_checksum.py
+++ b/arxiv/metadata/tests/test_checksum.py
@@ -1,0 +1,16 @@
+from arxiv.metadata.checksum import checksum_metadata
+
+def test_checksum_metadata():
+    class Md1:
+        def __init__(self):
+                self.title="bogus value"
+                self.authors="bogus value"
+                self.abstract="bogus value"
+                self.comments="bogus value"
+                self.report_num="bogus value"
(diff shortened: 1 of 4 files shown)


jonathanhyoung (approved):

arxiv/metadata/tests/test_checksum.py
```diff
@@ -0,0 +1,16 @@
...
+                self.authors="bogus value"
+                self.abstract="bogus value"
+                self.comments="bogus value"
+                self.report_num="bogus value"
+                self.journal_ref="bogus value"
+                self.doi="bogus value"
+                self.msc_class="bogus value"
+                self.acm_class="bogus value"
+
+    assert checksum_metadata(Md1())
```
carly-jones: this only tests that the function returns something that isn't `None` or `0`. suggest changing this to two test cases - passing two identical objects produces the same checksum, and passing two different objects produces different checksums

comment, not blocking

carly-jones (approved):

PR #396, opened 2026-03-04
CHECK-1310 add max len checks for all metadata; changes for ty check.

jonathanhyoung:


arxiv/metadata/tests/test_metacheck.py
```diff
@@ -791,7 +795,8 @@ def test_offsets():
     #
 
 def test_complaint2str():
-    assert complaint2str(-1) == "(Unknown issue)"
+    # This does not typecheck
```
bdc34: Probably just get rid of dead code and comments.

bdc34 (approved):

arxiv/metadata/tests/test_metacheck.py
jonathanhyoung: Done.

PR #398, opened 2026-03-25
SUBMISSION-80: License data updates

DavidLFielding:
This PR updates the license setting file in arxiv-base/licenses to support the new Submit 2.0 license page.

There are quite a few changes to the license metadata. Many labels and descriptions are updated per the final mockups of the license page (v5).

This metadata is used to build the license page (Step #3).

Submit 2.0:

LicenseChangesSubmit-2 0

Mockup version 5:

LicenseChanges-Mockupv5

THIS PR WILL NOT BE MERGED UNTIL WE ARE READY TO DEPLOY Submit 2.0.

### DIFF
--- a/arxiv/license/__init__.py
+++ b/arxiv/license/__init__.py
@@ -3,7 +3,6 @@ from typing import Optional
 
 """arXiv license definitions."""
 
-
 LICENSE_ICON_BASE_URI = '/icons/licenses'
 LICENSES = {
     # key is the license URI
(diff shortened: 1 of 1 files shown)


jweiskoff (approved):

jimentwood:
Just for my clarification.
Is a response required on the license selection page?
Or can a user proceed without selecting anything?

If it is required, can we see a screenshot of what is displayed when someone clicks Continue without having made a selection?

sdorphan:
Hi Jim,

I'm am reading this as clicking no radio button has the same effect as
selecting the "radio don't want to add an additional license" button (I may
be paraphrasing the option, not looking at the doc). Is that the case?
Personally I don't think it is a good idea to have two methods of getting
to the same place. One and one is active.

I think k we either need to require a selection or revert to removing the
"I choose nothing option"

jweiskoff (approved):

jimentwood:
It is fine to require selection and to have the initially proposed
"No Additional License".

I'm just curious on what the display is if someone clicks continue and gets some kind of warning.

sdorphan:
Some words wore missing in my original comment. "One is active and one is
passive."

PR #399, opened 2026-03-26
Config of servers in `arxiv.base.config.settings.URLS` fixed

bdc34:
In browse, setting `AUTH_SERVER` env var did not work to alter the host set for /login.

This is due to the value for `AUTH_SERVER` used in `URLS` being fixed at class definition time. The values in the list of tuples in `settings.URLS` never got updated with any new values if `setting.AUTH_SERVER` was later set with the value from the env var.

### DIFF
--- a/arxiv/base/config.py
+++ b/arxiv/base/config.py
@@ -42,49 +42,49 @@ Usually the same as BASE_SERVER but can be configured.
 HELP_SERVER = os.environ.get("HELP_SERVER", "info.arxiv.org")
 
 URLS = [
-    ("a11y", "/help/web_accessibility.html", HELP_SERVER),
-    ("about", "/about", HELP_SERVER),
-    ("about_give", "/about/give.html", HELP_SERVER),
-    ("about_people", "/about/people/index.html", HELP_SERVER),
-    ("acknowledgment", "/about/ourmembers.html", HELP_SERVER),
-    ("contact", "/help/contact.html", HELP_SERVER),
-    ("copyright", "/help/license/index.html", HELP_SERVER),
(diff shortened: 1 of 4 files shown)


arxiv/base/config.py
```diff
@@ -42,49 +42,49 @@
 HELP_SERVER = os.environ.get("HELP_SERVER", "info.arxiv.org")
```
ntai-arxiv, line 42: This is not used right?

ntai-arxiv (approved):
Rather than using _SERVER postfix, why not use '$' prefix and intern name with name[1:] ? so

`HELP_SERVER="$HELP_SERVER"` (or {} and use Python str.format() - after checking the first char $ or {)

and leave config.py, __init__ with small change - no need to quote HELP_SERVER by setting it to "{HELP_SERVER}" for example.

_server_name() -> _eval_app_config() and the setting's feature more universal.

arxiv/base/config.py
bdc34: This is used if someone uses the `arxiv.base.conifg` to assist in configuring their app. By default when 'arxiv.base.Base(app)' is run it is not used.

bdc34:
> Rather than using _SERVER postfix, why not use '$' prefix and intern name with name[1:] ? so
> 
> `HELP_SERVER="$HELP_SERVER"` (or {} and use Python str.format() - after checking the first char $ or {)
> 
> and leave config.py, **init** with small change - no need to quote HELP_SERVER by setting it to "{HELP_SERVER}" for example.
> 
> _server_name() -> _eval_app_config() and the setting's feature more universal.

I don't understand what change you'd like.

PR #401, opened 2026-04-04
[WOMBAT-396] OIDC PKCE support. Added

ntai-arxiv:
Ditch HTTPSimpleAuth in header and use PKCE, which is not only more secure, it also works even if the load balancer / reverse proxy stripping off some of headers as the secret is in the payload. Win-win-win-win. (no more simple auth, better security, better compatibility with the GCP infra, and newer standard.)

### DIFF
--- a/arxiv/auth/openid/oidc_idp.py
+++ b/arxiv/auth/openid/oidc_idp.py
@@ -1,6 +1,9 @@
 """
 OpenID connect IdP client
 """
+import base64
+import hashlib
+import secrets
 import urllib.parse
 from typing import List, Optional
 import requests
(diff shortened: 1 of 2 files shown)


bdc34:
The generation and verify of generation could be not in arxiv-base and just local code in /login + /callback. 

Nothing else is going to generate a pair for those endpoints.  The whole point is that nothing else should be able to generate these other than /login and /callback. At runtime, the goal is to never allow any one to forge these, even if they know how to do the code at https://github.com/arXiv/arxiv-base/pull/401/changes#diff-507b0214b3117da4467aeca56d41792ff883dcfdaaec4379a63352d2255f28baR28

bdc34 (approved):

PR #407, opened 2026-04-29
[WOMBAT-454] Let SQLAlchemy to assign the created/updated timestamps.…

ntai-arxiv:
…  MySQL can handle it but sqlite3 cannot. Also it seems like some don't have the DDL defined with timestamp.

While writing tests for Admin API, I noticed that "updated" is not updated. The test uses sqlite3, and mysql native timestamping does not work. With some tables lacking the timestamping, only sure way is to set the timestamp is to let SQLAlchemy's map column feature to set it. This may not be 100% backward compatible but I cannot come up with a better way to ensure the timestamp is always done. 

This may not be 100% cover all tables but ones important - documents, submissions, metadata are covered.

The terminal app `db_codegen_app.py` - I use it to run the codegen. It's a tool and not a part of actually using database.

### DIFF
--- a/arxiv/db/models.py-orig
+++ b/arxiv/db/models.py-orig
@@ -20,7 +20,7 @@ from typing import Optional, Literal, Any, Tuple, List
 import re
 import hashlib
 import datetime as dt
-from datetime import datetime, date
+from datetime import datetime, date, UTC
 from dateutil.tz import gettz, tzutc
 from sqlalchemy.dialects.mysql import VARCHAR
 from validators import url as is_valid_url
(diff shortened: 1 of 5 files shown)


arxiv/db/models.py-orig
```diff
@@ -68,6 +68,12 @@ from .column_types import Utf8Text, Utf8String
 tb_secret = settings.TRACKBACK_SECRET
 tz = gettz(settings.ARXIV_BUSINESS_TZ)
 
+def utc_now() -> datetime:
```
jonathanhyoung, line 71: Do you need to modify this file?

ntai-arxiv: yes. otherwise the merged code - models.py does not get this function. If you are interested in how this works, please read [the README.md](https://github.com/arXiv/arxiv-base/blob/develop/development/README.md).

> TL;DR - If you need to add to db/models.py, you either add it to db/models.py-orig or to db/arxiv-db-metadata.yaml.

jonathanhyoung (approved):

PR #408, opened 2026-04-30
Add publish of consecutive arXiv ID related columns to arXiv_submissions

norbusan:
Instead of using the submit_time for implementing consecutive orderering of submissions, we add two columns to submissions table.

- publish_anchor_time: where the override group lands. Computed as MAX(COALESCE(release_time, submit_time)) over the input set, so the group sits in the neighborhood of its latest member's natural slot - similar to the Perl code
- publish_order: within-group tiebreaker. A globally monotonic integer that prevents two operator calls colliding on the same anchor from interleaving by submission_id.

The publish processor's freeze sort key reads both columns: COALESCE(publish_anchor_time, release_time, submit_time), publish_order ASC, submission_id. For non-overridden rows both columns are NULL and the sort collapses to today's behavior. Only publish_order carries an index (sparse column, accelerates MAX(publish_order) for next-counter allocation).


arxiv/db/models.py
```diff
@@ -1149,6 +1149,8 @@ class Submission(Base):
     must_process: Mapped[Optional[int]] = mapped_column(Integer, server_default=FetchedValue())
     submit_time: Mapped[Optional[datetime]] = mapped_column(DateTime)
     release_time: Mapped[Optional[datetime]] = mapped_column(DateTime)
+    publish_anchor_time: Mapped[Optional[datetime]] = mapped_column(DateTime)
+    publish_order: Mapped[Optional[int]] = mapped_column(Integer, index=True)
```
jonathanhyoung, line 1153: Do you really need this to be indexed?

jonathanhyoung (approved):

arxiv/db/models.py
norbusan: Really need: no
But by far most of the entries in this column are NULL, and only for papers that require consecutive ordering there will be values.
That means, that the index will be very small as data.

What is your prime concern with adding this index?

jonathanhyoung: Every index adds a small amount of overhead. 

Thanks for the justification.

ntai-arxiv (approved):

PR #411, opened 2026-05-04
use modern timezone string

kyokukou:


norbusan (approved):
Yeah!

.github/workflows/pullrequest_tests.yaml
```diff
@@ -51,18 +51,6 @@ jobs:
...
-      - name: Install Firefox ESR and test driver
-        run: |
-          sudo add-apt-repository -y ppa:mozillateam/ppa
-          sudo apt update
-          sudo apt install -y firefox-esr
-          wget https://github.com/mozilla/geckodriver/releases/download/v0.36.0/geckodriver-v0.36.0-linux64.tar.gz
-          tar -xvzf geckodriver-v0.36.0-linux64.tar.gz
-          sudo mv geckodriver /usr/local/bin/
-          sudo chmod +x /usr/local/bin/geckodriver
-
```
norbusan, line 65: That change is not really something related to "use modern timezone string" ...

bdc34 (approved):

.github/workflows/pullrequest_tests.yaml
kyokukou: It is totally not related. It was blocking the build and Tai asked to remove it

PR #414, opened 2026-05-19
Fix the fastly purge's db code

ntai-arxiv:
(a) It could access the non-existing data before raising an exception.

(b) the db statement relied on not having "STRICT_TRANS_TABLES". (it works for sqlite3 as it is lax, and the prod db is okay with it) but, the query can and should select the document record first and then pick the abs_category so the STRICT_TRANS_TABLES does not come in play.

The test checks to see the query works with the option present.

Note that, the old code works for sqlite3 just because sqlite3 doesn't care about it. Since I was near (a), I fixed (b) as well. This does not work for the default installation of MySQL 8.4, fox example.

### DIFF
--- a/development/run_mysql_container.py
+++ b/development/run_mysql_container.py
@@ -23,26 +23,23 @@ def is_port_open(host: str, port: int) -> bool:
 def run_mysql_container(port: int, container_name: str ="mysql-test", db_name: str ="testdb",
                         root_password: str = "rootpassword") -> None:
     """Start a mysql docker"""
-    # mysql_image = "mysql:8.0.40"
-    mysql_image = "mysql:5.7.20"
+    mysql_image = "mysql:8.4"
 
     subprocess.run(["docker", "pull", mysql_image], check=True)
-    # subprocess.run(["docker", "kill", container_name], check=False)
     subprocess.run(["docker", "rm", container_name], check=False)
 
-    # for mysql mode, don't use ONLY_FULL_GROUP_BY
     argv = [
         "docker", "run", "-d", "--name", container_name,
         "-e", f"MYSQL_ROOT_PASSWORD={root_password}",
         "-e", "MYSQL_USER=testuser",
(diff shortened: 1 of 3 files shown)


development/run_mysql_container.py
```diff
@@ -23,26 +23,23 @@ def is_port_open(host: str, port: int) -> bool:
...
-        "-e", "MYSQL_SQL_MODE=STRICT_TRANS_TABLES,NO_ZERO_DATE,NO_ZERO_IN_DATE,ERROR_FOR_DIVISION_BY_ZERO",
         "-p", f"{port}:3306",
         mysql_image,
-        # Add this for mysql 8 and up to turn off SSL connection
-        # "--require_secure_transport=OFF"
-        "--sql-mode=STRICT_TRANS_TABLES,NO_ZERO_DATE,NO_ZERO_IN_DATE,ERROR_FOR_DIVISION_BY_ZERO"
+        "--require_secure_transport=OFF",
+        "--mysql-native-password=ON",
+        "--authentication_policy=mysql_native_password",
+        "--sql-mode=STRICT_TRANS_TABLES,NO_ZERO_DATE,NO_ZERO_IN_DATE,ERROR_FOR_DIVISION_BY_ZERO,ONLY_FULL_GROUP_BY"
```
bdc34, line 42: Are these the options being run in GCP?

bdc34 (approved):

development/run_mysql_container.py
ntai-arxiv: It's the default MySQL 8, which is more strict than prod MySQL 8 which is very lenient. Original implementation does not work with "STRICT_TRANS_TABLES" is on. If the test runs with it, it is definitely fine with the prod. IOW, the test is intending to be future proofing.

I understand the prod db came from old version of MySQL, but when there is an opportunity, even a small tech debt being removed, and proving it works is a plus.

PR #415, opened 2026-05-20
Add a basic arXiv holiday db table with created_at/by columns for audit

norbusan:


arxiv/db/arxiv_db_schema.sql
```diff
@@ -1071,6 +1071,20 @@ CREATE TABLE `arXiv_pilot_files` (
...
+-- Holiday calendar consumed by arxiv-publish-control-api's scheduler
+-- gate. Each row marks one date (America/New_York) on which no freeze
+-- and no announcement run. See publish git repo for the consumer.
+CREATE TABLE `arXiv_publish_holidays` (
+  `holiday_date` date NOT NULL,
+  `description` varchar(255) DEFAULT NULL,
+  `created_at` timestamp NOT NULL DEFAULT CURRENT_TIMESTAMP,
+  `created_by` varchar(64) DEFAULT NULL,
+  PRIMARY KEY (`holiday_date`)
+) ENGINE=InnoDB DEFAULT CHARSET=latin1;
```
bdc34: If it is a new table can we make the char set utf8mb4?

arxiv/db/arxiv_db_schema.sql
```diff
@@ -1071,6 +1071,20 @@ CREATE TABLE `arXiv_pilot_files` (
   CONSTRAINT `arXiv_pilot_files_cdfk3` FOREIGN KEY (`submission_id`) REFERENCES `arXiv_submissions` (`submission_id`)
 ) ENGINE=InnoDB AUTO_INCREMENT=1 DEFAULT CHARSET=latin1;
 /*!40101 SET character_set_client = @saved_cs_client */;
+DROP TABLE IF EXISTS `arXiv_publish_holidays`;
+/*!40101 SET @saved_cs_client     = @@character_set_client */;
+/*!50503 SET character_set_client = utf8mb4 */;
+-- Holiday calendar consumed by arxiv-publish-control-api's scheduler
+-- gate. Each row marks one date (America/New_York) on which no freeze
+-- and no announcement run. See publish git repo for the consumer.
+CREATE TABLE `arXiv_publish_holidays` (
```
bdc34: Jim suggested we change the name of the table and I agree with Jim on this. Can we call it `arXiv_freeze_skip_days` or something similar? The day isn't when there is no publish, but when there is no freeze. This is often a source of confusion.

arxiv/db/arxiv_holidays_seed.sql
```diff
@@ -0,0 +1,73 @@
+-- Seed for arXiv_publish_holidays.
+--
+-- Source: arxiv-lib.git/lib/arXiv/Config/Holidays.pm (@HOLIDAYS array,
+-- lines 38-95 as of 2026-05). One row per ISO date. The `description`
+-- column carries the inline perl comment verbatim.
+--
+-- After loading this seed, ops must keep the perl @HOLIDAYS array and
+-- this table in sync (dual-write) until the perl side migrates onto
+-- the DB. See the operational rule in publish.git CLAUDE.md.
+
```
bdc34, line 12: Since the existing perl code is the full list of holidays going back to 1991. So we should have that data somewhere and we'll throw out the Perl file sooner or later.

arxiv/db/arxiv_db_schema.sql
bdc34: Or maybe change the column `holiday_date` to `freeze_skip_date`?

bdc34: Actually let's name the table `arXiv_holidays` and name the column `freeze_skip_date`.

bdc34 (changes requested):
I've asked for some changes. Let me know if they don't make sense.

arxiv/db/arxiv_db_schema.sql
norbusan: I am fine with that, I was just wondering that I faintly remember that we had some hacks around utf-8 in latin1 tables that are bound to some pydantic models. I might mis-remember, though.
If that is the case, does utf8mb4 work anyway?

arxiv/db/arxiv_db_schema.sql
norbusan: Ok, I have done that. One question: The model name would now be `Holiday` I guess, that is what I have made.

arxiv/db/arxiv_holidays_seed.sql
norbusan: Where do you see 1991-2019? The arxiv-lib Holidays.pm contains the same values as I have here:
https://github.com/arXiv/arxiv-lib/blob/develop/lib/arXiv/Config/Holidays.pm#L38-L95

arxiv/db/arxiv_db_schema.sql
bdc34: Yes, utf8mb4 should work fine because we are not reading or writing from the arxiv-lib Perl code which has the unhealthy encoding code.

arxiv/db/arxiv_db_schema.sql
bdc34: I'm fine with `arXiv_holidays`, I think it is important to name the column `freeze_skip_date` since that is what it has meant in the past. 

This was discussed on slack with Jim and he said he wanted the UI to not call it holidays but didn't put a constraint on what the table name would be. So these are fine.

arxiv/db/arxiv_holidays_seed.sql
bdc34: If you go back in the git history of that file there are older holidays. It got cleared out at the start of this year.

bdc34 (approved):

bmaltzan (approved):

PR #418, opened 2026-06-01
Move metadata title checks into the qa package

carly-jones:
## Description
Move metadata title checks into the QA package
  
  Refactors the QA check framework and migrates title validation into the qa package. All other metadata field checks will be moved in a subsequent PR.

## Changes
  Base class changes (base.py):
  - Switched BaseCheck from ClassVar attributes to instance-level attributes
  - Added BaseGenericCheck for checks that target a specific data + field pair
  - Added BaseGenericPatternCheck for regex-based sub-checks (pattern → offsets)
  - Added BaseAggregateCheck for composite checks that run multiple sub-checks and aggregate results; failures only fail the aggregate when the sub-check's disposition is REJECT
  
  New models (models.py):
  - Disposition enum (OK / WARN / REJECT) — controls whether a failure is blocking
  - Inputs as the typed data container passed into every check 
  - Metadata, MetadataProtocol, Result, and Offset pydantic models

  New check (metadata/title.py):
  - ValidTitleCheck — aggregate check with 17 sub-checks covering emptiness, length, casing, whitespace, brackets, HTML, control characters, encoding, and TeX patterns; empty title
  rejects, all other checks warn


qa/qa/checks/base.py
```diff
@@ -1,34 +1,174 @@
...
+        return {
+            "disposition": self.disposition,
+            "data": self.data,
+            "field": self.field,
+            "pattern": self._pattern,
+        }
+
+    def _run(self, inputs: Inputs) -> Result:
+        """The pattern applied to the configured field should not return any matches."""
+        v = getattr(getattr(inputs, self.data), self.field)
```
jonathanhyoung, line 121: Do these calls need defaults values?

qa/qa/checks/models.py
```diff
@@ -0,0 +1,87 @@
...
+    journal_ref: str | None
+    doi: str | None
+    msc_class: str | None
+    acm_class: str | None
+
+
+class Inputs(BaseModel):
+    """Data dependencies for checks."""
+
+    fulltext: str | None = None
```
jonathanhyoung, line 82: Fulltext can be really large.

jonathanhyoung (approved):
Are you using ruff for formatting? I thought I ran my code through ruff, so I'm surprised there were so many edits to metacheck.py .

carly-jones:
> Are you using ruff for formatting? I thought I ran my code through ruff, so I'm surprised there were so many edits to metacheck.py .

ah I didn't mean for the formatter to make changes to metacheck - I'll remove that from this PR. yes, I am using ruff - for this particular case I extended the line length to make the aggregate checks easy to read. there is a make command (`make format`) to make it easy

qa/qa/checks/base.py
carly-jones: BaseGenericPatternCheck inherits from BaseGenericCheck, which requires these, so we will never be in a situation where they are not available.

I specifically made most aspects of the checks mandatory (disposition, data, field) to prevent misconfiguration issues. there shouldn't be a check that we add without them

if we need to relax that later, we can

qa/qa/checks/models.py
carly-jones: yeah it is possible that we retype this to be a reference to the fulltext (i.e. url or such) at some point, we will see. for now this is just a placeholder

PR #421, opened 2026-06-05
[QA-76] Move and refactor remaining metadata checks

carly-jones:
This PR adds aggregate checks for all remaining submission metadata fields (authors, abstract, comments, report_num, journal_ref, doi, msc_class, acm_class) to the `qa` package. Check ids and failure messages are aligned with the `Complaint` IntEnum and `complaint2str` in `arxiv/metadata/`. All test cases are covered. 

Aggregate checks (each metadata field) now have ids 500-590 (incremented by 10).

This PR also updates the current disposition field attached to each check (static configuration) to be an "on_failure_policy", and returns a disposition as the end state of running that check on an input.


lauren-arxiv:
The code overall looks good to me! I'm a little confused/concerned about a few high-level things (some of which actually aren't in this PR).

1. Aggregate check failure policy isn't actually used logically AFAICT? Maybe it won't matter in practice but I wonder if the "lack of failure policy" can be made explicit for the aggregate check template.

2. I'm quite confused about the numbering of check ID's -- I see that aggregate checks for each field are given ID's 500:10:590, but how does it combine with generic ID's for specific checks on specific fields (e.g. author not empty)? My understanding of the initial plan was that a generic check like NotEmpty (check_id = 1) would get additively combined with a field offset (title offset = 500) for something like TitleNotEmpty to have check_id = 501. TitleNotTooShort would have check_id 502, etc. Under that scheme, I imagine the offset would just be passed as an extra parameter when instantiating the generic check for a specific field.

3. Ultimately check ID's don't really matter as long as they're 1-to-1, but I'm concerned that check ID's are going to collide. In fact `id = 15` is used twice (there might be other examples, I didn't check all of them). Is there a scheme underlying the ID's as presented here?

4. I'm not in love with offsets & recursive check results being part of *every* single check result -- including, eventually, non-metadata-validation checks? Although flagged-terms checks do use offsets IIRC, and kinda act like aggregate checks, so maybe it'll be useful to keep them around as a vestigial field. Just wondering if this will get baked in to all checks, and how that affects design for non-metadata-validation checks.

5. Are we against naming the Result model something slightly more descriptive, like CheckResult? Not a huge deal but I worry about name collisions when imported into other systems.

6. The logic overall seems good and is very thoughtfully designed. I'm wondering if we can write a little documentation about some of the design decisions (even just in the directory's README) sooner rather than later, though I know it's not 100% hammered out yet. (Happy to help with this too)

lauren-arxiv (approved):

qa/qa/checks/generic/author_name.py
```diff
@@ -0,0 +1,202 @@
...
+    """A check that parses authors and validates individual author entries."""
+
+    @abstractmethod
+    def _check_author(self, keyname: str, firstname: str, suffix: str) -> bool:
+        """Return True if the author passes, False if it fails."""
+        ...
+
+    def _run(self, inputs: Inputs) -> Result:
+        v = getattr(getattr(inputs, self.data), self.field)
+        for author in parse_author_affil(v):
```
jonathanhyoung, line 47: Seems as though this will be called once for each author check - ?

carly-jones: as discussed we think this is not going to be a bit sink - we can change it later if we see performance issues

carly-jones:
confirming here that we met and discussed @lauren-arxiv 's concerns. we are good to move forward

in sum:
1. this was a bug and has been fixed
2. check ids are nested, not combined (e.g. generic checks always have their own number, not a combined one with the aggregate check running them). this is now clarified by returning the full config with results
3. these were taken from the previous implementation, but I will double check that we have no repeats
4. these fields are included for typing reasons, and they may in fact be used later (offsets are relevant to all text related checks)
5. this is to avoid a naming collision with the CheckResults table (which should really be renamed CheckFailures) - with the import hierarchy we should never have namespace collisions with any other objects
6. documentation will be updated on this PR

carly-jones:
3. there were some repeats added by Claude when the author name checks were moved. they have been renumbered to 51-56
6. a README has been added

lauren-arxiv (approved):

PR #428, opened 2026-06-23
QA-148 Full text checks

jonathanhyoung:
Ticket: https://arxiv-org.atlassian.net/browse/QA-148

Implemented MissingTextCheck and VeryShortTextCheck.

https://arxiv-org.atlassian.net/browse/QA-173 Implemented OverlapCheck
https://arxiv-org.atlassian.net/browse/QA-174 Implemented WithdrawalCheck

### DIFF
--- a/qa/qa/checks/fulltext/text_checks.py
+++ b/qa/qa/checks/fulltext/text_checks.py
@@ -1,7 +1,7 @@
 
 from qa.checks.base import BaseCheck
 
-from qa.checks.models import OnFailurePolicy, QaDataRegistry, Metadata, Result
+from qa.checks.models import OnFailurePolicy, QaDataRegistry, Result
 
 
 class MissingTextCheck(BaseCheck):
@@ -37,6 +37,7 @@ class VeryShortTextCheck(BaseCheck):
     description = "The full text extracted is too short."
     on_failure_policy = OnFailurePolicy.REJECT
     failure_message = "Very short text (< 1400 words). Check full text."
+    
     required_inputs = {"fulltext"}
 
     def _run(self, data_registry: QaDataRegistry) -> Result:
--- a/qa/qa/checks/fulltext/text_checks.py
+++ b/qa/qa/checks/fulltext/text_checks.py
@@ -42,7 +42,7 @@ class VeryShortTextCheck(BaseCheck):
 
     def _run(self, data_registry: QaDataRegistry) -> Result:
         fulltext = data_registry.fulltext
-        if fulltext is None or fulltext is "":
+        if fulltext is None or fulltext == "":
             # Problem: we should only report the problem with missing texts
             passed = True
             return self._result(passed)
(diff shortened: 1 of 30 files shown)


carly-jones:
we initially scoped this to just a single check because the two feel redundant. can we remove missing text?

qa/qa/checks/fulltext/text_checks.py
```diff
@@ -0,0 +1,57 @@
...
+
+from qa.checks.models import OnFailurePolicy, QaDataRegistry, Result
+
+
+class MissingTextCheck(BaseCheck):
+    """Check for missing text (probably PDF invalid or password protected)."""
+
+    name = "missing_text"
+    display_name = "Missing Text"
+    id = 14
```
carly-jones, line 12: these ids seem to be repeats (see checks/generic/text.py)

qa/qa/checks/fulltext/text_checks.py
```diff
@@ -0,0 +1,57 @@
...
+            passed = False
+            return self._result(passed, message = self.failure_message)
+        
+
+class VeryShortTextCheck(BaseCheck):
+    """Check for very short text."""
+
+    name = "short_text"
+    display_name = "Short Text"
+    id = 15
```
carly-jones, line 36: these ids seem to be repeats (see checks/generic/text.py)

carly-jones:
thinking about this - we already have a generic check `NotTooShort`. would it be cleaner just to update the fulltext model and use that check?

e.g.

```
class Fulltext(BaseModel):
      content: str | None = None

class QaDataRegistry(BaseModel):
      fulltext: Fulltext | None = None
      ...

class FulltextNotTooShort(BaseAggregateCheck):
      name = "fulltext_not_too_short"
      display_name = "Fulltext Not Too Short"
      id = {something free}
      version = "1.0.0" 
      description = "Fulltext meets basic length requirements."
      on_failure_policy = OnFailurePolicy.REJECT
      failure_message = "Fulltext is too short."
      required_data = {"fulltext"}
      
      _checks = (
          NotTooShort(1400, on_failure_policy=OnFailurePolicy.REJECT,
                      data="fulltext", field="content"),
      )
```

this way the configuration (min 1400) will also be returned with the results and registry. what do you think?

qa/qa/checks/fulltext/text_checks.py
jonathanhyoung: Maybe we should detect that automatically?

Where are you getting your IDs?  I'm using the ones which were originally assigned as check_ids - those already have meaning elsewhere in the system.

jonathanhyoung: Ids renumbered.

qa/qa/checks/fulltext/text_checks.py
jonathanhyoung: metadata checks renumbered.

carly-jones:
as discussed in slack - no need to update this. just remove MissingTextCheck as it is covered by `_validate_inputs` (and also redundant with VeryShort).

also a minor note - all the checks are named in the positive (like most form validation packages) so it would be nice to name this `FulltextNotTooShort`

jweiskoff (approved):

PR #429, opened 2026-06-25
[WOMBAT-575] jwt expiration

ntai-arxiv:


norbusan (approved):
Looks reaasonable, but I didn't check the math.

Isn't there a shared library for these kind of RFC handling? I cant imagine that we are the first who have to do that?

ntai-arxiv:
> Looks reaasonable, but I didn't check the math.
> 
> Isn't there a shared library for these kind of RFC handling? I cant imagine that we are the first who have to do that?

I asked the same Q. I looked into pyjwt and couldn't find any. I might be able to find one in the wild but I couldn't bring myself to introduce another library just for this structure.

PR #437, opened 2026-07-15
QA-241 drop FK relationship from arXiv_check_results to arXiv_checks.

jonathanhyoung:


ntai-arxiv (approved):
I have done my very best to warn this is a bad idea.

PR #438, opened 2026-07-16
[WOMBAT-606] User session start time fix

ntai-arxiv:


arxiv/auth/legacy/sessions.py
```diff
@@ -154,13 +154,21 @@ def create(authorizations: domain.Authorizations,
...
     logger.debug('create session for user %s', user.user_id)
-    start = datetime.now(tz=UTC)
-    end = start + timedelta(seconds=util.get_session_duration())
+    # iat - "Issued at" - the time at which the token was issued.
+    # util.epoch() rounds, which can land ~0.5s in the future. start_time later becomes a token's iat
+    # (via generate_cookie -> util.epoch), and a future iat trips PyJWT's not-yet-valid check (ImmatureSignatureError).
+    # Floor start to whole seconds so epoch() is exact and the issued time never leads the clock;
+    # start_epoch and end both derive from this same floored instant.
+    start = datetime.now(tz=UTC).replace(microsecond=0)
+    start_epoch = int((start - datetime.fromtimestamp(0, tz=util.EASTERN)).total_seconds())
```
bdc34: This seems a little strange. Use of EASTERN seem odd with epoch and then going back to UTC. 

Consider something like
```
start = datetime.now(tz=timezone.utc).replace(microsecond=0)
start_epoch = int(start.timestamp()) 
end = start + timedelta(seconds=util.get_session_duration())
```

bdc34 (approved):
Strange use of timezones but seems to work.

Consider using the minor suggested change.

bdc34 (changes requested):
Please get rid of the poetry.lockfile and pyprojec.toml changes

The types are dev time use only?

Package management on the production arxiv-auth can be a pain and I don't want to get involved in it for dev dependencies.

bdc34 (approved):

PR #439, opened 2026-07-20
QA-71 separate lists of metadata and fulltext checks.

jonathanhyoung:


qa/qa/checks/__init__.py
```diff
@@ -30,6 +30,11 @@
...
     OversizeCheck(),
     WithdrawalCheck(),
+]
+
+fulltext_checks: list[BaseCheck] = [
     FulltextExtractedCheck(),
     FulltextNotTooShortCheck(),
 ]
+
+all_checks = set(metadata_checks).union(set(fulltext_checks))
```
carly-jones: sets are unordered. can we keep `all_checks` as a list, and just add the two together? we have uniqueness covered by the namespace as well as the unit tests you wrote

carly-jones: also this should be named `checks` (that'll resolve your type error)

carly-jones: I'm ok with renaming `checks` to `all_checks` here but that change will also have to be made in the service (it imports `checks`)

carly-jones:
I renamed `metadata_checks` to `submission_checks` (as discussed in slack)

carly-jones:
there is something up with the diversion of branches on this PR, so I'm closing this and will open a fresh one from develop

PR #440, opened 2026-07-20
Do not install base dependencies in QA

carly-jones:
This PR fixes the dependency override for the `qa` package so that all of the base package deps are not installed. This was previously done but buggy - apparently a version pin is required. This now shrinks the list of installed packages from ~1150 to 18.

This PR also runs `make format` over all `qa` package files and adds a format check to the `lint-test-package.yml` workflow to enforce formatting.


jonathanhyoung (approved):
I love -1000-line commits!

PR #441, opened 2026-07-21
[QA-189] Update text extraction failed check

carly-jones:
This PR updates the TextExtractionSuccessful check based on the failure report produced by pdftotext. It also updates the FulltextNotTooShort check to return configuration. Both checks now have a convenience class method.


qa/qa/checks/fulltext/extraction.py
```diff
@@ -0,0 +1,38 @@
+from qa.checks.base import BaseCheck
+
+from qa.checks.models import FulltextReport, OnFailurePolicy, QaDataRegistry, Result
+
+
+class TextExtractionSuccessful(BaseCheck):
+    name = "text_extraction_successful"
+    display_name = "Text Extraction Successful"
```
jonathanhyoung, line 8: This is the field used for display on the left-hand side of the Check UI. Can we come up with a better display name, so the EUST knows what to look for when this check fails?

qa/qa/checks/fulltext/structure.py
```diff
@@ -0,0 +1,39 @@
+from qa.checks.base import BaseCheck
+
+from qa.checks.models import OnFailurePolicy, QaDataRegistry, Result
+
+
+class FulltextNotTooShort(BaseCheck):
+    name = "fulltext_not_too_short"
+    display_name = "Fulltext Not Too Short"
```
jonathanhyoung, line 8: This display name will not help the EUST determine what action to take when the test fails.

qa/qa/checks/fulltext/extraction.py
jweiskoff: We should only display a result when a check fails. It's fine to include text indicating a success, but the check shouldn't display in the modal in Check if it succeeds.

qa/qa/checks/fulltext/structure.py
jweiskoff: Can we rename this to remove the double-negative and then make it make more sense?

qa/qa/checks/fulltext/extraction.py
carly-jones: All of the checks extending BaseCheck have names/display names in the positive - this is a form validation convention, i.e. "check that text extraction is successful". This is valuable for consistency and readability of the code. If failed, the result in the UI will show the failure message, which makes it very clear what the problem is.

This will be reviewed in UAT and can easily be adjusted if it isn't clear

qa/qa/checks/fulltext/structure.py
carly-jones: See my comment above - again, this is checking that the fulltext is not too short. This is existing convention in form/data validation.

The failure message is clear and will display on failure

jonathanhyoung: How is the failure message going to be propagated to the EUST Check UI?

carly-jones: I will work with the rest of the Check team when we are ready to release these, but the failure message should display with the result in the LHP

carly-jones: added in refinement today: https://arxiv-org.atlassian.net/browse/QA-308

jonathanhyoung (approved):
Approved, but I think we are kicking the can down the road here.

PR #442, opened 2026-07-23
Update qa policies and models, bugfixes

carly-jones:
This PR separates Metadata and SubmitEventInfo models and updates submission checks, including their policies.

It also bugfixes one of the check classes attributes (should be `required_data` not `required_inputs`) as well as the broken unique id tests which were not being recognized by pytest because of the file name. Repeat check ids (14) were also fixed.


qa/qa/checks/models.py
```diff
@@ -77,10 +79,18 @@ class FulltextReport(BaseReport):
...
 
 
-class SubmissionMetadata(BaseModel):
+class SubmitEventInfo(BaseModel):  # check which fields are always provided by the snapshot and which are optional
+    """Information about the submission provided by the submit event."""
+
+    type: Literal["new", "rep", "wdr", "jref", "cross"]
+    is_oversize: bool
+    data_version: int
+    metadata_version: int
```
jonathanhyoung, line 88: Why did these move off (Submission)Metadata ? they can come from the database as well, right?

carly-jones: https://github.com/arXiv/arxiv-qa/pull/194
In this PR, the big change is that we are fetching `Metadata` as well as the other submit event info  (`SubmitEventInfo`) from the pubsub message rather than from a store when the service detects a pubsub trigger. In the case that it isn't a pubsub, `Metadata` is fetched from the db, and the submit event info is not available.

If we have a place to fetch this submit event info other than the pubsub message, we can update this later. as of now, I don't think we do.

jonathanhyoung (approved):

PR #448, opened 2026-07-31
QA-164 submitter name in authors

jonathanhyoung:
Also includes code for  QA-176 hold all HTML submissions.

### DIFF
--- a/qa/qa/checks/metadata/authors.py
+++ b/qa/qa/checks/metadata/authors.py
@@ -1,6 +1,9 @@
 """Author metadata checks."""
 
-from qa.checks.base import BaseAggregateCheck
+import re
+from arxiv.authors import parse_author_affil
+
+from qa.checks.base import BaseCheck, BaseAggregateCheck, EmptyFieldError
 from qa.checks.models import QaDataRegistry, OnFailurePolicy, Metadata, Result
 from qa.checks.generic.text import (
     AllBracketsBalanced,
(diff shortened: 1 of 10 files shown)


carly-jones:
did you see the code at `qa/checks/generic/author_name.py`? I think some of the helpers in here are duplicative logic

carly-jones:
if we can't use existing logic here, I'd prefer that we move any helpers into the check(s) themselves so they can be properly versioned, e.g. minimize functions outside of checks.

jonathanhyoung:
> did you see the code at `qa/checks/generic/author_name.py`? I think some of the helpers in here are duplicative logic

This PR checks whether the *submitter's name* appears in the list of authors.

The checks in author_name are detecting invalid author names, specifically:
* Names of LLMs
* Lone surnames
* Semicolons, brackets, numbers, affiliations

(We should add variations on anonymous to this list.)

There is a small amount of overlapping logic which might be pulled out (the check for whether the authors list is a collaboration), but it is different in the two checks.

qa/qa/checks/metadata/authors.py
```diff
@@ -73,3 +76,121 @@ def check(cls, authors: str | None) -> Result:
...
+
+    @property
+    def config(self) -> dict:
+        return {
+            **super().config,
+            "failure_flag_id": self.failure_flag_id,
+        }
+
+    def _run(self, data_registry: QaDataRegistry) -> Result:
+        if data_registry.metadata is None:
```
carly-jones, line 154: checks for required data aren't needed here - see BaseCheck

qa/qa/checks/metadata/authors.py
```diff
@@ -73,3 +76,121 @@ def check(cls, authors: str | None) -> Result:
...
+    display_name = "Authors Contains Submitter Name"
+    id = 30
+    version = "1.0.0"
+    description = "Submitter name found in authors field."
+    on_failure_policy = OnFailurePolicy.WARN
+    failure_message = "Submitter name not found in authors field (proxy check)."
+
+    required_data = {"metadata", "submit_event_info"}
+
+    failure_flag_id = "submitter-name-not-found-in-authors-field"
```
carly-jones: I think this can be removed

qa/qa/checks/metadata/authors.py
```diff
@@ -73,3 +76,121 @@ def check(cls, authors: str | None) -> Result:
...
+                    any_long_name1_found = True
+                    break
+                
+        return any_long_name1_found or (all_name1_names_short and all_short_name1_names_found)
+
+    @property
+    def config(self) -> dict:
+        return {
+            **super().config,
+            "failure_flag_id": self.failure_flag_id,
```
carly-jones: same as above - I think this can be removed. we aren't reading flags from a report in this check

qa/qa/checks/metadata/authors.py
```diff
@@ -73,3 +76,121 @@ def check(cls, authors: str | None) -> Result:
...
+
+        authors = data_registry.metadata.authors
+
+        submitter_name = data_registry.submit_event_info.submitter_name
+        submitter_names = AuthorsContainsSubmitterName.normalize_author_name(submitter_name).split()
+
+        passed = False
+
+        if authors is None:
+            return self._result(passed=True)
```
carly-jones, line 167: authors should never be empty - perhaps raise an EmptyFieldError instead? you could also override the _validate_data method to check the field, similar to BaseGenericCheck

carly-jones (approved):
approved with a few suggestions

qa/qa/checks/metadata/authors.py
carly-jones: only line 92

qa/qa/checks/metadata/authors.py
carly-jones: only line 159

carly-jones:
Looks like you need to run `make format`

qa/qa/checks/metadata/authors.py
jonathanhyoung: It is used at line 159 by the config function.

qa/qa/checks/metadata/authors.py
jonathanhyoung: Why not just remove the config completely?

qa/qa/checks/metadata/authors.py
jonathanhyoung: We want only one error to be identified if the authors field is empty, but what is going to prevent this check from running when it is empty?

qa/qa/checks/metadata/authors.py
jonathanhyoung: The type checker gives an error if I do not check here.

PR #449, opened 2026-08-03
[QA-251] Add registry of metadata checks

carly-jones:
This PR adds a registry of just metadata checks for use in the ModAPI metadata validation endpoint. 

It also exposes the field as a class attribute for each aggregate check, enforces instantiating generic checks with named params, and cleans up imports of generic checks by exposing them in the init file.


qa/qa/checks/__init__.py
```diff
@@ -17,7 +17,7 @@
 from qa.checks.fulltext.extraction import TextExtractionSuccessful  # noqa
 from qa.checks.fulltext.structure import FulltextNotTooShort  # noqa
 
-submit_event_checks: list[BaseCheck] = [
+metadata_checks: list[BaseCheck] = [
```
jonathanhyoung, line 20: This is going to get confusing. 

There are checks we want to run at submit time which we do not want to run on the metadata before then, or which we do not want to disclose to the submitter when they fail (e.g. proxy checks).

carly-jones: I'm not sure I understand your comment. what is your suggestion/edit?

jonathanhyoung: Multiple lists for different purposes.

carly-jones: what is your suggestion specifically? this PR gives us three lists: metadata_checks, submit_event_checks, and checks (the full registry)

jonathanhyoung: K I see what you did now.

jonathanhyoung (approved):

PR #451, opened 2026-08-03
QA-312 Abstract must be in English.

jonathanhyoung:


carly-jones:
Two initial comments -
- The `langdetect` package doesn't seem to be in active development. How did we decide on this one? Do we have other choices that are actively maintained?
- I don't think it will suffice to not run language detection <30 chars. There are hints on the package README for alternative approaches - use a threshold on the probabilities returned, for example. We should take into account whether we are holding these for review or blocking them.

jonathanhyoung:
> Two initial comments -
> 
> * The `langdetect` package doesn't seem to be in active development. How did we decide on this one? Do we have other choices that are actively maintained?

Language detection is basically a well-understood technology. Also, the library I added, https://pypi.org/project/fasttext-langdetect/, was last updated in May this year.

> * I don't think it will suffice to not run language detection <30 chars. There are hints on the package README for alternative approaches - use a threshold on the probabilities returned, for example. We should take into account whether we are holding these for review or blocking them.

I just benchmarked this library against 100k of our abstracts, of which only 51 were shorter than 30 characters. The scores were all over the place, but only one was incorrect (french). But we have to check for min chars (5) or else "not english" triggers on a very short abstract, which isn't reasonable.

carly-jones (approved):
Approved with two comments:
- It would be nice to have the configuration you added (min_chars) returned with the config
- It would be nice to have the check covered by tests outside of the abstract aggregate check tests

PR #456, opened 2026-08-14
[QA-317] Update metadata checks

carly-jones:
This PR updates `qa` to match desired metadata checking behavior outlined in [this confluence page](https://arxiv-org.atlassian.net/wiki/spaces/QA/pages/2475393025/Metadata+Check+Build+List+2026-08-03).

It also splits generic text checks into several files for discoverability, keeping the same import statement (all checks are exposed at the top level).


qa/qa/checks/generic/content.py
```diff
@@ -0,0 +1,127 @@
...
+
+class DoesNotContainBareDoi(BaseGenericPatternCheck):
+    name = "does_not_contain_bare_doi"
+    display_name = "Does Not Contain Bare DOI"
+    id = 10040
+    version = "1.0.0"
+    description = "The value does not contain a bare DOI number (e.g. 10.1234/abc)."
+    failure_message = "Contains a DOI."
+
+    _pattern = r"(?i)^[0-9][0-9].[0-9]+/[^ ]*$"
```
jonathanhyoung: This pattern should be r"[0-9]+(\.[0-9]+)*/.*".

qa/qa/checks/generic/doi.py
```diff
@@ -0,0 +1,31 @@
...
+    name = "doi_has_valid_format"
+    display_name = "DOI Has Valid Format"
+    id = 10050
+    version = "1.0.0"
+    description = "Each space-separated DOI in the value matches the expected DOI format."
+    failure_message = failure_message = (
+        "Contains a DOI that doesn't match the expected format: digits, followed by a period, followed by '/' and a suffix (e.g. '10.1234/abc123')."
+    )
+
+    _pattern = r"[0-9]+(\.[0-9]+)/.*"
```
jonathanhyoung: The pattern should be r"[0-9]+(\.[0-9]+)*/.*"

qa/qa/checks/generic/whitespace.py
```diff
@@ -0,0 +1,58 @@
...
+    display_name = "No Unnecessary Space in Parens"
+    id = 10033
+    version = "1.0.0"
+    description = "The value does not contain leading or trailing spaces immediately inside parentheses."
+    failure_message = "Unnecessary space inside parentheses."
+
+    _pattern = r"\(\s|\s\)"
+
+
+class DoesNotContainSpaceAfterOpenParen(BaseGenericPatternCheck):
```
jonathanhyoung: Remove, covered by NoUnnecessarySpaceInParens.

qa/qa/checks/generic/prefixes.py
```diff
@@ -0,0 +1,47 @@
...
+
+class DoesNotBeginWithAbstract(BaseGenericPatternCheck):
+    name = "does_not_begin_with_abstract"
+    display_name = "Does Not Begin With Abstract"
+    id = 10005
+    version = "1.0.0"
+    description = "The value does not begin with the literal prefix 'abstract' or 'abstract:'."
+    failure_message = "Begins with 'abstract'."
+
+    _pattern = r"(?i)^abstract:?\b"
```
jonathanhyoung: I think we should be checking for "Abstract:" here, not "Abstract". I bet we have paper titles which start with "Abstract algebra ..."

qa/qa/checks/generic/content.py
carly-jones: from slack: 
"
Chapter and verse (from https://www.doi.org/doi-handbook/DOIHandbook_2025.pdf page 38):

The DOI prefix is composed of a directory indicator optionally followed by a registrant code. Where a registrant code is present, the two components are separated by a U+002E FULL STOP.

EXAMPLE 1: In the DOI name 10.5594/SMPTE.ST2067-21.2020, “10.5594” is the DOI prefix, “10” is the directory indicator and “5594” is the registrant code.

The directory indicator consists of a sequence of digits between 0 and 9. It is usual-ly equal to “10” but other directory indicators may be used in the future by the DOI Foundation. The registrant code consists of sequences of digits between 0 and 9, separated by U+002E FULL STOP.

EXAMPLE 2: In the hypothetical “10.500.100” DOI prefix, the directory indicator is “10” and the registrant code is “500.100”.

So the regex I recommend is r"[0-9]+(\.[0-9]+)*/.*"
"

qa/qa/checks/generic/doi.py
carly-jones: from slack: 
"
Chapter and verse (from https://www.doi.org/doi-handbook/DOIHandbook_2025.pdf page 38):

The DOI prefix is composed of a directory indicator optionally followed by a registrant code. Where a registrant code is present, the two components are separated by a U+002E FULL STOP.

EXAMPLE 1: In the DOI name 10.5594/SMPTE.ST2067-21.2020, “10.5594” is the DOI prefix, “10” is the directory indicator and “5594” is the registrant code.

The directory indicator consists of a sequence of digits between 0 and 9. It is usual-ly equal to “10” but other directory indicators may be used in the future by the DOI Foundation. The registrant code consists of sequences of digits between 0 and 9, separated by U+002E FULL STOP.

EXAMPLE 2: In the hypothetical “10.500.100” DOI prefix, the directory indicator is “10” and the registrant code is “500.100”.

So the regex I recommend is r"[0-9]+(\.[0-9]+)*/.*"
"

qa/qa/checks/generic/content.py
carly-jones: (note that markdown is dropping the asterisks)

qa/qa/checks/generic/doi.py
carly-jones: (note that markdown is dropping the asterisks)

qa/qa/checks/generic/content.py
carly-jones: `r"[0-9]+(\.[0-9]+)*/.*"`

qa/qa/checks/generic/doi.py
carly-jones: `r"[0-9]+(\.[0-9]+)*/.*"`

qa/qa/checks/generic/content.py
dginev: Their team (CrossRef) also has a recommendation:

https://www.crossref.org/blog/dois-and-matching-regular-expressions/

carly-jones: @jonathanhyoung please take a look at that ref and lmk what you think ^^

qa/qa/checks/generic/structure.py
```diff
@@ -0,0 +1,95 @@
...
+
+        for index, char in enumerate(v):
+            if char in bracket_pairs:
+                stack.append((char, index))
+            elif char in ")}]":
+                if stack and bracket_pairs[stack[-1][0]] == char:
+                    stack.pop()  # bracket closed, remove from stack
+                else:
+                    error_index = index
+                    break
```
jonathanhyoung, line 61: This code rejects intervals like "(0,1]".

qa/qa/checks/generic/whitespace.py
```diff
@@ -0,0 +1,58 @@
...
+
+class NoExtraWhitespace(BaseGenericPatternCheck):
+    name = "no_extra_whitespace"
+    display_name = "No Extra Whitespace"
+    id = 10025
+    version = "1.0.0"
+    description = "The value does not contain multiple consecutive spaces, trailing whitespace before a newline, or irregular comma spacing."
+    failure_message = "Excessive or irregular whitespace."
+
+    _pattern = r"\s+\n|[^ \t\n,][ \t][ \t]+[^ \t\n,]|\s+,(\s*,)*[a-zA-Z]?|\s*,(\s*,)+"
```
jonathanhyoung: This pattern also alerts on ",," - I don't think calling that "excessive whitespace" is going to help our submitters.

Suggestion: check for ",," and "<nospace>,<nospace>" elsewhere.

jonathanhyoung (approved):

qa/qa/checks/generic/structure.py
carly-jones: this produces a warning on abstract, comments, and title and only rejects on authors. it functions as intended

qa/qa/checks/generic/content.py
jonathanhyoung: Crossref is offering regexps which aggressively match DOIs. We want to do the opposite: we want to identify DOIs but not match on things which could realistically be a journal reference, such as "2026/AAAI". 

If you want to go in this direction, I recommend falling back on looking for the 10.<id> pattern, which is more likely to be a DOI than anything else, viz:

r"/^10.\d{4,9}/.+" 

This will fail to detect uncommon DOIs, but still be useful.

qa/qa/checks/generic/whitespace.py
carly-jones: done. comma cases covered by DoesNotContainSpaceBeforeComma and DoesNotContainUnspacedComma

qa/qa/checks/generic/doi.py
jonathanhyoung: Actually, I'd like to change my recommendation to r"[0-9]+(\.[0-9]+)*/.+". I don't think the suffix can realistically be empty, although the spec is not explicit on the subject.

